PR #10, opened 2026-01-18
Turret and climb subsystems

Fatblabs:
This is only turret subsystem

### DIFF
--- a/src/main/java/frc/robot/Subsystems.java
+++ b/src/main/java/frc/robot/Subsystems.java
@@ -20,6 +20,7 @@ import frc.robot.subsystems.SpindexerSubsystem;
 import frc.robot.subsystems.TurretSubsystem;
 import frc.robot.subsystems.VisionSubsystem;
 import frc.robot.subsystems.drivebase.CommandSwerveDrivetrain;
+import frc.robot.subsystems.drivebase.DrivebaseWrapper;
 
 public class Subsystems {
   public static class SubsystemConstants {
--- a/src/main/java/frc/robot/subsystems/drivebase/DrivebaseWrapper.java
+++ b/src/main/java/frc/robot/subsystems/drivebase/DrivebaseWrapper.java
@@ -1,4 +1,4 @@
-package frc.robot.subsystems;
+package frc.robot.subsystems.drivebase;
 
 import com.ctre.phoenix6.Utils;
 import edu.wpi.first.math.Matrix;
@@ -10,7 +10,6 @@ import edu.wpi.first.math.kinematics.SwerveDriveKinematics;
 import edu.wpi.first.math.kinematics.SwerveModulePosition;
 import edu.wpi.first.math.numbers.N1;
 import edu.wpi.first.math.numbers.N3;
-import frc.robot.subsystems.drivebase.CommandSwerveDrivetrain;
 import java.util.function.Supplier;
 
 /**


src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,171 @@
...
+import edu.wpi.first.wpilibj2.command.FunctionalCommand;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Hardware;
+
+public class TurretSubsystem extends SubsystemBase {
+  // ------ VARIABLES ------//
+  private final TalonFX motor1;
+  private MotionMagicVoltage request;
+
+  private final int GEAR_RATIO = 9000; // It's over 9000
```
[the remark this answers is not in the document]
Fatblabs: Of course

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,236 @@
...
+  private void moveMotor(double targetDegrees) {
+    motor1.setControl(request.withPosition(Units.degreesToRotations(targetDegrees)));
+  }
+
+  private double calculateTargetDegrees() {
+    // 158.84 Inches in the Y Direction
+    // 182.11 inches in the X Direction
+
+    SwerveDriveState driveState = m_driveTrain.getState(); // Get current drive state
+    Pose2d currentPose = driveState.Pose;
```
jamesdooley4: FYI that this isn't the location of the turret itself, which will be offset from the center of the robot by some amount. There's a helper in another branch to calculate the turret location from the robot pose, so please add a comment starting with `TODO: ` to update this to use the future turret pose helper

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,236 @@
...
+  private final TalonFX motor1;
+  private MotionMagicVoltage request;
+
+  private final int GEAR_RATIO = 9000; // It's over 9000
+  private final double MAXIMUM_DERIVATIVE = 0.3;
+  private final double TURRET_MIN = -90.0; // In Degrees
+  private final double TURRET_MAX = 90.0; // In Degrees
+
+  private static final double HUB_X_INCHES = 182.11;
+  private static final double HUB_Y_INCHES = 158.84;
```
jamesdooley4: This should now be calculatable from the AprilTag poses in the AprilTagFieldLayout for the 2026 field. Though somewhere common we should specify which of the two fields to use (welded or not welded), since that will change depending on what field location our robot is playing on.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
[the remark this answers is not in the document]
jamesdooley4: Something like 30 or 60 are more reasonable gear ratios.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,171 @@
...
+
+    var MotionMagicConfig = configs.MotionMagic;
+    MotionMagicConfig.MotionMagicCruiseVelocity = 1.5;
+    MotionMagicConfig.MotionMagicAcceleration = 160;
+    MotionMagicConfig.MotionMagicJerk = 1000;
+
+    configs.Feedback.SensorToMechanismRatio = GEAR_RATIO;
+
+    motor1.getConfigurator().apply(configs);
+  }
```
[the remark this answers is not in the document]
jamesdooley4: May do in the future as part of a common motor tuning helper. Not going to do now.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,236 @@
...
+            previousTimeStamp[0] = Timer.getTimestamp();
+            previousCurrent[0] = motor1.getTorqueCurrent(true).getValueAsDouble();
+            hits[0] = 0;
+            motor1.setControl(new VoltageOut(0.3));
+          },
+          () -> {
+            double now = Timer.getTimestamp();
+            double currentNow = motor1.getTorqueCurrent(true).getValueAsDouble();
+            double currentDT = now - previousTimeStamp[0];
+            if (currentDT < MIN_DT) {
```
jamesdooley4: Should be the absolute value of `currentDT` since getTorqueCurrent() can return both positive & negative numbers and the delta might be a large negative number.

jamesdooley4 (changes requested):

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,269 @@
...
+    turretRotationFieldRelative =
+        TurretNetworkTable.getDoubleTopic("Turret Rotation field relative").publish();
+    robotRotation = TurretNetworkTable.getDoubleTopic("Robot rotation").publish();
+    errorDeg = TurretNetworkTable.getDoubleTopic("errorDeg").publish();
+    targetDegrees = TurretNetworkTable.getDoubleTopic("Target Degrees").publish();
+    current = TurretNetworkTable.getDoubleTopic("Current").publish();
+  }
+
+  private void moveMotor(double targetDegrees) {
+    motor1.setControl(request.withPosition(Units.degreesToRotations(targetDegrees)));
```
markpete: Is there a way or is it worth considering knowing when the turret is moving/aligning, so that we don't fire unless it's within some measure of "target aligned"?

Also, can a rotation be interrupted and updated with a new position?  I can imagine the bot moving around and the turret needing to stop a current update to go to a new one.  Otherwise, the turret might lag too far behind the position updates.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,269 @@
...
+
+  private double calculateTargetDegrees() {
+    // 158.84 Inches in the Y Direction
+    // 182.11 inches in the X Direction
+
+    SwerveDriveState driveState = m_driveTrain.getState(); // Get current drive state
+    Pose2d currentPose = driveState.Pose;
+    Rotation2d rotation = currentPose.getRotation();
+
+    double robotRotation = rotation.getDegrees();
```
markpete: Can you just get both of these in Radians, then you don't have to convert back and forth?  Honestly, it might be simpler to just work in Radians.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,236 @@
...
+    double TurretRotation = motor1.getPosition().getValueAsDouble() * 360;
+    double TurretRotationsRad =
+        MathUtil.angleModulus(Units.degreesToRadians(robotRotation + TurretRotation));
+    double TurretRotationFieldRelative = Units.radiansToDegrees(TurretRotationsRad);
+
+    Translation2d hub =
+        new Translation2d(Units.inchesToMeters(HUB_X_INCHES), Units.inchesToMeters(HUB_Y_INCHES));
```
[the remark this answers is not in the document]
markpete: Should we change the magic numbers to meters and just work in that coordinate system?  Is there specific value in having them in inches in your code?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,269 @@
...
+    Translation2d difference = hub.minus(currentPose.getTranslation());
+    double requiredAngles =
+        Units.radiansToDegrees(
+            Math.atan2(
+                difference.getY(),
+                difference.getX())); // This will give the angle from the difference of x and y.
```
markpete: I'm not sure your comment is correct.  Doesn't atan2 calc the angle in radians from the positive x-axis to the line that passes through the X,Y point specified?  What is your coordinate in your working system for the x-axis?  Is it 0 or 90 degrees?  Assuming it's 0 based on your comment, so -pi would be to the left and +pi would be the right.  Is that correct?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,269 @@
...
+            if (!interrupted) {
+              motor1.setPosition(Units.degreesToRotations(TURRET_MAX + 0.5));
+              motor1.setControl(request.withPosition(Units.degreesToRotations(TURRET_MAX)));
+              readyToShoot = true;
+            }
+          },
+          () -> (hits[0] >= MIN_HITS),
+          TurretSubsystem.this);
+    }
+    motor1.setPosition(Units.degreesToRotations(TURRET_MAX + 0.5));
```
markpete: Does this mean that we can only zero when pointing to the right?  Would it make sense to zero left or right depending on which is closer?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,269 @@
...
+            case IDLE:
+            default:
+              stop();
+              break;
+          }
+        })
+        .finallyDo(interrupted -> stop());
+  }
+
+  public void manualMove(DoubleSupplier joystick) {
```
markpete: Do we need to consider the left side (-90) as well as the right side?

vendordeps/photonlib.json
```diff
@@ -0,0 +1,71 @@
+{
+    "fileName": "photonlib.json",
+    "name": "photonlib",
+    "version": "v2026.1.1-rc-2",
```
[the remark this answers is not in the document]
markpete: https://github.com/PhotonVision/photonvision/releases

src/main/java/frc/robot/subsystems/TurretSubsystem.java
jamesdooley4: Steamnasium is AndyMark, not welded.

vendordeps/photonlib.json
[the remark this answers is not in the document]
koolpoolo: no just delete it since it has no use in the code and we aren't using it this year

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,255 @@
...
+  final DoublePublisher current;
+
+  // PLACEHOLDER VALUE. This will probably be handled elsewhere
+  private boolean readyToShoot = false;
+  private static turretState currentState = turretState.IDLE;
+
+  // --- CONSTRUCTOR --- //
+  public TurretSubsystem(SwerveDrivetrain drivetrain) {
+    // --- MOTOR SETUP ---//
+    m_turretMotor = new TalonFX(Hardware.TURRET_MOTOR_ID);
```
markpete: TURRET_MOTOR_ID is undefined.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,255 @@
...
+    request = new MotionMagicVoltage(0);
+    configureMotors();
+    // --- Hardware --- //
+    hub = AllianceUtils.getHubTranslation2d();
+    m_driveTrain = drivetrain;
+    turretTransform =
+        new Transform2d(new Translation2d(TURRET_X_OFFSET, TURRET_Y_OFFSET), Rotation2d.kZero);
```
markpete: No need to create the Translation2d object, just create the Transform2d directly.

`turretTransform = new Transform2D(TURRET_X_OFFSET, TURRET_Y_OFFSET, Rotation2d.kZero);`

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,255 @@
...
+    } // trying to go further +
+    if (rad <= TURRET_MIN && volts < 0) {
+      stop();
+      return;
+    } // trying to go further -
+    m_turretMotor.setControl(new VoltageOut(volts));
+  }
+
+  public void stop() {
+    m_turretMotor.setControl(new VoltageOut(0));
```
markpete: This gets called about every 20ms.  So you're creating a new VoltageOut() object every 20ms, which will balloon and cause the garbage collector to do a ton of work.  Create it outside of the loop and only do it once.

markpete: Recommend you make it a class private constant and re-use it wherever you need it.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,255 @@
...
+    currentState = turretState.AUTO;
+  }
+
+  public static void MANUAL() {
+    currentState = turretState.MANUAL;
+  }
+
+  // --- MOTOR CONFIGS --- //
+  private void configureMotors() {
+    TalonFXConfiguration configs = new TalonFXConfiguration();
```
markpete: Manual vs. Hardware Limits: You manually check limits in 
manualMove
 (Lines 186-193) and clamp the target in 
calculateTargetRadians
 (Lines 114-116). However, you rely on the PID controller in 
moveMotor
 to respect the clamp. If the PID overshoots (e.g., due to high velocity or kF), the turret could hit the physical hard stops violently because the motor controller itself doesn't know the limits.
Recommendation: Configure Firmware Soft Limits in 
configureMotors
 using TalonFXConfiguration. This pushes the safety check to the motor controller (1kHz loop), which is much safer and faster than the RIO (50Hz loop).

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,255 @@
...
+    Slot0Configs slot1 = configs.Slot0;
+
+    slot1.kS = 0.25; // Required voltage to overcome static friction.
+    slot1.kV = 2; // 2 volts to maintain 1 rps
+    slot1.kA = 4; // 4 volts required to maintain 1 rps/s
```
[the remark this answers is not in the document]
Fatblabs: resolved

jamesdooley4:
@Fatblabs the formatting & simulation checks are both failing. Please take a look & fix the errors they found.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,171 @@
...
+
+    configs.MotorOutput.NeutralMode = NeutralModeValue.Brake;
+    m_climb_motor.getConfigurator().apply(configs);
+  }
+
+  // command that if scheduled would automatically align with the tower
+  public Command autoAlignRoutine(ClimbState state) {
+    return Commands.run(
+            () -> {
+              if (currentState == ClimbState.L0 && isClose()) {
```
markpete: We should probably add a isClimbing variable.  I believe this gets run every loop cycle (20ms) and it could get scheduled multiple times or while running.  Adding a !isClimbing to the check here would prevent it getting scheduled multiple times.

markpete (commented):
Some updates to consider.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,261 @@
...
+import frc.robot.Hardware;
+import frc.robot.util.AllianceUtils;
+import java.util.function.DoubleSupplier;
+
+public class TurretSubsystem extends SubsystemBase {
+  // ------ VARIABLES ------//
+  private final TalonFX m_turretMotor;
+  private MotionMagicVoltage request;
+
+  private static final double GEAR_RATIO = 1.0 / 10.0; // It's over 9000
```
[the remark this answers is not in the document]
Fatblabs: resolved

vendordeps/Phoenix6-26.1.1.json
jamesdooley4: Curious why the file was renamed back to the version # based one, instead of the version independent name supplied when visiting the URL here.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,261 @@
...
+  final DoublePublisher current;
+
+  // --- CONSTRUCTOR --- //
+  public TurretSubsystem(SwerveDrivetrain drivetrain) {
+    // --- MOTOR SETUP ---//
+    m_turretMotor = new TalonFX(Hardware.TURRET_MOTOR_ID);
+    request = new MotionMagicVoltage(0);
+    configureMotors();
+    // --- Hardware --- //
+    hub = AllianceUtils.getHubTranslation2d();
```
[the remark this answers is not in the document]
Fatblabs: assuming that we're always targeting the hub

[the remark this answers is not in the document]
iamawesomecat: > I think I did this wrong, I think It should have been hub = new Pose2d(AllianceUtils.getHubTranslation2d(), Rotation2d.kZero).

In theory yes, but the thing is, alliance is set after code boots, because it isnt set until after you connect to the FMS of the field.

src/main/java/frc/robot/Controls.java
```diff
@@ -32,7 +33,19 @@ public class Controls {
...
+  private static final int TURRET_TEST_CONTROLLER_PORT_2 = 4;
 
+  // Cool enums
+
+  // --- Turret STATES ---- //
+  public static enum TurretState {
+    MANUAL,
+    AUTO,
+    IDLE
+  }
```
iamawesomecat: Does this need to be in controls.java?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,252 @@
...
+  // ------- AUTOZERO ------ //
+  private static final double MIN_VELOCITY = 0.3;
+  private static final double MIN_HITS = 5;
+
+  // ----- HARDWARE OBJECTS ----- //
+  private final SwerveDrivetrain m_driveTrain;
+  private final Pose2d hub;
+  private final Transform2d turretTransform;
+
+  private boolean zeroed = false;
```
markpete: The zeroed flag seems correct but ensure it's reset if the subsystem is disabled or re-initialized if needed.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,152 @@
+package frc.robot.subsystems;
```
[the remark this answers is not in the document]
iamawesomecat: I think its commented so he can test other things without having to debug.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,347 @@
...
+import java.util.function.DoubleSupplier;
+import java.util.function.Supplier;
+
+public class TurretSubsystem extends SubsystemBase {
+  // ------ VARIABLES ------//
+  private final TalonFX m_turretMotor;
+  private MotionMagicVoltage request;
+
+  private static final double GEAR_RATIO = 20;
+  // private static final double TURRET_MAX = Units.degreesToRadians(90);
```
markpete: Recommend you remove dead code.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,347 @@
...
+    slot0.kD = k_D;
+
+    slot0.kS = k_S;
+    slot0.kV = k_V;
+    slot0.kA = k_A;
+
+    // ---CURRENT LIMITS --- //
+    configs.CurrentLimits.StatorCurrentLimitEnable = true;
+    configs.CurrentLimits.SupplyCurrentLimitEnable = true;
+    configs.CurrentLimits.StatorCurrentLimit = 25;
```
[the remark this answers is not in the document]
markpete: You're doing the right order below in 302-305.

src/main/java/frc/robot/Controls.java
```diff
@@ -234,6 +254,61 @@ private void configureLauncherBindings() {
     launcherTuningController.y().onTrue(s.flywheels.setVelocityCommand(60));
   }
 
+  private void configureTurretBindings(boolean isXbox) {
```
markpete: Instead of a boolean, you might make this an enum of
```
enum CongrollerType {
XBOX, PS2;
}
```

iamawesomecat:
I might be dumb and not understand how things work,but when you change the min or max of the turret (which we will have to do) it stops working. Either fix that or add a explanation why

src/main/java/frc/robot/subsystems/VisionSubsystem.java
```diff
@@ -116,23 +123,87 @@ private void processLimelight(
...
+        // System.out.println("updated time");
+      }
+    }
+  }
 
-        drivebaseWrapper.addVisionMeasurement(
+  private void processLimelightmt2(
+      LLCamera camera, StructPublisher<Pose3d> rawFieldPoseEntry, RawFiducial rf) {
+
+    disableVision = SmartDashboard.getBoolean("Disable Vision", false);
```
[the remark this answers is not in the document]
Fatblabs: I was told to merge Viggo's vision subsystem with my branch.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,347 @@
...
+
+  public Command manualMove(DoubleSupplier joystick) {
+    return Commands.run(
+        () -> {
+          if (!zeroed) {
+            m_turretMotor.setControl(ZERO_VOLTS);
+            return;
+          }
+          double cmd = MathUtil.applyDeadband(joystick.getAsDouble(), 0.10);
+          double volts = cmd * 2.0;
```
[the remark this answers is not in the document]
Fatblabs: Ive tested it with 2 and it works. 5 volts is pretty fast i plan on using that for my autozero routine

PR #18, opened 2026-01-27
Autos

Jetblackdragon:


src/main/java/frc/robot/Robot.java
```diff
@@ -119,7 +138,18 @@ public void disabledPeriodic() {}
...
   /** This autonomous runs the autonomous command selected by your {@link RobotContainer} class. */
   @Override
-  public void autonomousInit() {}
+  public void autonomousInit() {
+    if (AutoLogic.getSelectedAuto() != null) {
+      if (Robot.isSimulation()) {
+
+
+        fuelSimulation.clearFuel();
+        fuelSimulation.spawnStartingFuel();
```
markpete: We clearFuel() and spawnStartingFuel() in the constructor.  Do we need it here too?

src/main/java/frc/robot/util/FuelSim.java
```diff
@@ -0,0 +1,623 @@
...
+    double impulse = 0.5 * (1 + FUEL_COR) * (b.vel.minus(a.vel).dot(normal));
+    double intersection = FUEL_RADIUS * 2 - distance;
+    a.pos = a.pos.plus(normal.times(intersection / 2));
+    b.pos = b.pos.minus(normal.times(intersection / 2));
+    a.addImpulse(normal.times(impulse));
+    b.addImpulse(normal.times(-impulse));
+  }
+
+  private static void handleFuelCollisions(ArrayList<Fuel> fuels) {
+    for (int i = 0; i < fuels.size() - 1; i++) {
```
markpete: Is this routine auto-generated?  It's an O(N^2) algorithm, so will be slow.

Issue: The collision detection algorithm is O(N²).

spawnStartingFuel creates approximately 400 fuel objects.
stepSim is called 5 times per robot loop (subticks = 5).
Total operations: 400² pairs * 5 subticks * 50 loops/sec ≈ 40,000,000 checks per second. Impact: This will likely cause severe simulation lag, overruning the 20ms loop time, and making the simulation unusable on standard hardware. 

Remediation:
Implement spatial partitioning (e.g., a hash grid or quadtree) to reduce collision checks to O(N).
Significantly reduce the number of fuel particles simulated (e.g., only simulate fuel near the robot).
Reduce subticks count if physics stability permits.

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
markpete: Issue: commandsMap has a size of 1. The loop condition i < 1 (starting at i=1) fails immediately. Impact: The "Auto Mode" chooser will only ever show "0". Remediation: Fix the loop range or properly populate commandsMap with the intended data.

src/main/java/frc/robot/Robot.java
SnappleRamen: I removed this. Thanks for finding that oversight

src/main/java/frc/robot/Robot.java
```diff
@@ -46,7 +48,24 @@ protected Robot() {
...
+          0.8,
+          0.7,
+          () -> subsystems.drivebaseSubsystem.getState().Pose,
+          () -> subsystems.drivebaseSubsystem.getState().Speeds);
+      fuelSimulation.registerIntake(0.4, 0.8, 0.4, 0.8, () -> true);
+
+      fuelSimulation.start();
+
+      fuelSimulation.clearFuel();
+      fuelSimulation.spawnStartingFuel();
```
jamesdooley4: This was already done on line 54. Is it needed here? Presumably the .clearFuel also isn't needed since this is a brand new instance.

PR #24, opened 2026-01-31
Intake subsystem

cwkpete7383:
Created the Intake Subsystem code. Although it's unfinished, I'm creating this so that comments can be added to it. Update 2:00 pm: it's finished yippee

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -220,11 +220,13 @@ public class Controls {
     driverController
         .rightTrigger()
         .whileTrue(
-            Commands.sequence(
-                    AutoAim.autoAim(s.drivebaseSubsystem, s.hood, s.flywheels),
-                    Commands.parallel(
-                        s.spindexerSubsystem.startMotor(), s.feederSubsystem.startMotor()))
-                .withName("Autorotate, Autoaim done, feeder and spindexer started"))
+            Commands.parallel(
+                s.turretSubsystem.rotateToHub(),
+                Commands.sequence(
+                        AutoAim.autoAim(s.drivebaseSubsystem, s.hood, s.flywheels),
+                        Commands.parallel(
+                            s.spindexerSubsystem.startMotor(), s.feederSubsystem.startMotor()))
+                    .withName("Autorotate, Autoaim done, feeder and spindexer started")))
         .toggleOnFalse(
             Commands.parallel(
                 s.hood.hoodPositionCommand(0.0), s.flywheels.setVelocityCommand(0.0)));
@@ -235,6 +237,7 @@ public class Controls {
                     s.hood.zeroHoodCommand(),
                     s.turretSubsystem.zeroTurret(),
                     s.intakePivot.zeroPivot())
+                .withName("Zero subsystems")
                 .ignoringDisable(true));
     if (s.flywheels.TUNER_CONTROLLED) {
       launcherTuningController


src/main/java/frc/robot/Subsystems.java
```diff
@@ -33,13 +40,17 @@ public Subsystems() {
       drivebaseSubsystem = null;
       drivebaseWrapper = new DrivebaseWrapper();
     }
+    if (INTAKEROLLERS_ENABLED && INTAKEPIVOT_ENABLED){
+      intakeSubsystem = new IntakeSubsystem(true, true);
+    }
```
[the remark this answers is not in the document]
cwkpete7383: The condition is meant to be restrictive. If even one of the components of the intake isn't enabled (especially the pivot), then it won't work.

src/main/java/frc/robot/Subsystems.java
```diff
@@ -46,6 +52,12 @@ public Subsystems(Mechanism2d mechanism2d) {
       drivebaseSubsystem = null;
       drivebaseWrapper = new DrivebaseWrapper();
     }
+    if (INTAKEROLLERS_ENABLED && INTAKEPIVOT_ENABLED) {
```
jamesdooley4: Should be `||` (or) so the intake subsystem can be created on the Alpha bot that has rollers but no pivot

src/main/java/frc/robot/Controls.java
```diff
@@ -182,6 +183,13 @@ private void configureAutoAlignBindings() {
     driverController.rightBumper().whileTrue(FuelAutoAlign.autoAlign(this, s));
   }
 
+  private void configureIntakeBindings() {
+    if (s.intakeSubsystem == null) {
+      return;
+    }
+    driverController.b().whileTrue(s.intakeSubsystem.runIntake(1));
```
jamesdooley4: Now that the pattern has been started, please add a new intake test controller & use that here.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,213 @@
...
+      1000; // change this value and ln 44 if u wanna change position
+  private final DoubleTopic leftRollerTopic;
+  private final DoubleTopic rightRollerTopic;
+  private final DoublePublisher leftRollerPub;
+  private final DoublePublisher rightRollerPub;
+  private final FlywheelSim leftRollerSim;
+  private final FlywheelSim rightRollerSim;
+  private final SingleJointedArmSim pivotSimV2;
+  LinearSystem rollerSystem = LinearSystemId.createFlywheelSystem(DCMotor.getKrakenX60(1), 1, 1);
+  //   LinearSystem pivotSystem = LinearSystemId.createElevatorSystem(DCMotor.getKrakenX60(1), 40,
```
markpete: Please delete unused code.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,213 @@
...
+    var pivotMotionMagicConfigs = talonFXConfigs1.MotionMagic;
+    pivotMotionMagicConfigs.MotionMagicAcceleration = 3000;
+    pivotMotionMagicConfigs.MotionMagicJerk = 10;
+
+    pivotMotor.getConfigurator().apply(talonFXConfigs1);
+  }
+
+  // rollers configs
+  private void TalonFXRollerConfigs() {
+    var talonFXConfigs2 = new TalonFXConfiguration();
```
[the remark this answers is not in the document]
markpete: I see you have a talonFXConfigs1 above, but it's a local variable inside of a private function, so you won't have naming conflicts here, since they are each in their own private scopes.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,213 @@
...
+        () -> {
+          pivotMotor.setControl(
+              pivot_request1.withPosition(
+                  PIVOT_DEPLOYED_POS)); // placeholder value, change during testing
+          leftRollers.set(speed);
+          rightRollers.setControl(
+              new Follower(
+                  Hardware.INTAKE_MOTOR_ONE_ID,
+                  MotorAlignmentValue.Opposed)); // opposite direction as left rollers
+          System.out.println(
```
markpete: Do you need this print out?  Should it instead be published to the driver dashboard?  I suspect this gets called frequently and this System.out() could create a lot of noise.

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -0,0 +1,213 @@
...
+            RadiansPerSecond.of(pivotSimV2.getVelocityRadPerSec() * PIVOT_GEAR_RATIO)
+                .in(RotationsPerSecond));
+  }
+
+  @Override
+  public void periodic() {
+    leftRollerPub.set(leftRollers.getVelocity().getValueAsDouble());
+    rightRollerPub.set(rightRollers.getVelocity().getValueAsDouble());
+
+    pivotArm.setAngle(Units.radiansToDegrees(pivotSimV2.getAngleRads()));
```
markpete: NullPointerException:

**Issue**: pivotSimV2 is only initialized when RobotBase.isSimulation() is true (lines 83-102). On a real robot, pivotSimV2 will be null.
**Impact:** The code will crash immediately upon initialization/enabling on the RoboRIO.
**Fix:** Wrap the simulation-only logic in an if (RobotBase.isSimulation()) block or null check.

markpete (changes requested):
Off to a good start.  Good job!  There are a few things to clean up before this is ready to check-in.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,233 @@
...
+    var slot0Configs = talonFXConfigs.Slot0;
+    talonFXConfigs.MotorOutput.NeutralMode = NeutralModeValue.Brake;
+
+    // pivot configs
+
+    slot0Configs.kS = 0.9;
+    slot0Configs.kV = 0;
+    slot0Configs.kA = 0;
+    slot0Configs.kP = 40;
+    slot0Configs.kI = 5;
```
iamawesomecat: kI is heavily taboo in frc. I know you copied this from last year, but just letting you know

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,233 @@
...
+    slot0Configs.kD = 0;
+    slot0Configs.kG =
+        0.048; // change PID values during testing, these are placeholders from last year's robot
+
+    var pivotMotionMagicConfigs = talonFXConfigs.MotionMagic;
+    pivotMotionMagicConfigs.MotionMagicAcceleration = 3000;
```
iamawesomecat: sorry 3000 rotations per second per second?😭

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,233 @@
...
+          leftRollers.set(speed);
+        },
+        () -> {
+          leftRollers.stopMotor();
+        });
+  }
+
+  @Override
+  public void simulationPeriodic() {
+    if (leftRollerSim == null) {
```
iamawesomecat: Do not need to check if sim is null, as simulation periodic will only run when running sim, which is also what causes you to create leftRollerSim

src/main/java/frc/robot/subsystems/Intake/IntakeSim.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private final MechanismRoot2d pivotShoulder;
+    private final MechanismLigament2d pivotArm;
+
+    private static final double PIVOT_GEAR_RATIO = (54.0 / 12.0) * (54.0 / 18.0) * (48.0 / 18.0);
+    private static final double ARM_LENGTH_METERS = Units.inchesToMeters(10.919);
+
+    LinearSystem rollerSystem = LinearSystemId.createFlywheelSystem(DCMotor.getKrakenX60(1), 1, 1);
+
+    public IntakeSim(Mechanism2d mechanism2d, TalonFX leftRollers, TalonFX rightRollers) {
```
iamawesomecat: pass in pivot as well, unless you're planning on making a seperate pivotSim?

iamawesomecat (changes requested):

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,145 @@
...
+import edu.wpi.first.networktables.NetworkTableInstance;
+import edu.wpi.first.wpilibj.RobotBase;
+import edu.wpi.first.wpilibj.smartdashboard.Mechanism2d;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.Commands;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Hardware;
+
+public class IntakeSubsystem extends SubsystemBase {
+  private static final double speed2 = 1;
```
markpete: Unclear variable naming.  speed2 implies a speed1 exists, which it does not.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,145 @@
...
+  private final TalonFX pivotMotor;
+  private final TalonFX leftRollers; // one
+  private final TalonFX rightRollers; // two
```
markpete: [nit]  The // one and // two comments aren't helpful here.  I'd remove them.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,145 @@
...
+    System.out.println("configure rollers");
+  }
+
+  public Command runIntake() {
+    return Commands.runEnd(
+        () -> {
+          System.out.println("left roll go"); // for testing
+          leftRollers.set(speed2);
+          rightRollers.setControl(followerRequest); // opposite direction as left rollers
+          pivotMotor.setControl(pivotRequest.withPosition(PIVOT_DEPLOYED_POS));
```
markpete: This runs every 20ms. While Phoenix 6 setControl is designed to be called frequently, leftRollers.set usually calls the setter. For mostly static values (like constant speed), this is unnecessary overhead, though often negligible unless the bus is saturated. Correction: Consider startEnd if values don't change, or verify that repeated calls are intended for safety.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,145 @@
...
+
+    leftRollers.getConfigurator().apply(rollerConfigs);
+    rightRollers.getConfigurator().apply(rollerConfigs);
+    System.out.println("configure rollers");
+  }
+
+  public Command runIntake() {
+    return Commands.runEnd(
+        () -> {
+          System.out.println("left roll go"); // for testing
```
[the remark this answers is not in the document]
cwkpete7383: These print statements are going to be removed for the final code, I added them there for debugging. Should I put these as values in NetworkTables anyway?

src/main/java/frc/robot/subsystems/Intake/IntakeSim.java
```diff
@@ -0,0 +1,111 @@
...
+import edu.wpi.first.wpilibj.simulation.RoboRioSim;
+import edu.wpi.first.wpilibj.simulation.SingleJointedArmSim;
+import edu.wpi.first.wpilibj.smartdashboard.Mechanism2d;
+import edu.wpi.first.wpilibj.smartdashboard.MechanismLigament2d;
+import edu.wpi.first.wpilibj.smartdashboard.MechanismRoot2d;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+public class IntakeSim extends SubsystemBase {
+  private static final double updateSim = 0.02;
```
markpete: @Jetblackdragon - Should we move the updateSim value of 20ms to a common file instead of having each *Sim.java set this locally?  That way we can configure it in a single place.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+  public Trigger atPosition(double position) {
+    return new Trigger(() -> Math.abs(getPivotPos() - position) < POS_TOLERANCE);
+  }
+  public Command deployIntake() {
+    return Commands.runOnce(
+      () -> {
+        System.out.println("pivot position: " + pivotPos);
+        System.out.println(getPivotPos() == PIVOT_RETRACTED_POS);
+        if (getPivotPos() == PIVOT_RETRACTED_POS) {
```
markpete: **Risky Floating Point Comparison** 
`if (getPivotPos() == PIVOT_RETRACTED_POS) {`
Issue: Comparison of doubles using == is prone to failure due to floating point precision. If getPivotPos() is off by even a tiny epsilon, this check fails, preventing deployment.

Recommendation: Use a tolerance check.

`if (Math.abs(getPivotPos() - PIVOT_RETRACTED_POS) < POS_TOLERANCE) {`

src/main/java/frc/robot/subsystems/Intake/IntakeSim.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private final MechanismRoot2d pivotRoot;
+  private final MechanismLigament2d pivotArm;
+
+  private static final double PIVOT_GEAR_RATIO = 36.0;
+  private static final double ROLLERS_GEAR_RATIO = 1.0;
+  private static final double ARM_LENGTH_METERS = Units.inchesToMeters(10.919);
+  private static final double ARM_START_POS = 15.0;
+
+  LinearSystem rollerSystem = LinearSystemId.createFlywheelSystem(DCMotor.getKrakenX60(1), 1, 1);
```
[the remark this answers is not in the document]
cwkpete7383: Not sure how to calculate the MOI, but these rollers work fine in simulation and irl.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,232 @@
...
+    return runOnce(
+        () -> {
+          pivotMotor.stopMotor();
+        });
+  }
+
+  public Command deployIntake() {
+    return Commands.runOnce(
+        () -> {
+          System.out.println(getPivotPos() == PIVOT_RETRACTED_POS);
```
markpete: Remove this line.  System.out.printlin() is slow and can lag bot.  Also, this appears to be debug only and doesn't give enough information, so good to remove.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,232 @@
...
+          if (getPivotPos() == PIVOT_RETRACTED_POS) {
```
markpete: Is getPivotPos() guaranteed to be exact or is there a range of correctness within an epsilon?  e.g. PIVOT_DEPLOYED_POS +/- 0.05;  Since it is a double even being off a tiny amount due to precision issues would cause this to fail. Here's an alternative way you could code this...

```
double targetPos = (getPivotPos() < 0.0) ? PIVOT_DEPLOYED_POS : PIVOT_RETRACTED_POS;
pivotMotor.setControl(pivotRequest.withPosition(targetPos);
```

markpete: Alternataively, you could do a check similar to what you do for atPosition();
`Math.abs(getPivotPos() - position) < POS_TOLERANCE)`

If you do re-use this, maybe consider breaking it out into a private helper method within the class.

src/main/java/frc/robot/subsystems/Intake/IntakeSubsystem.java
```diff
@@ -0,0 +1,232 @@
...
+    return runOnce(
+        () -> {
+          pivotMotor.setControl(pivotRequest.withPosition(pos));
+          targetPos = pos;
+        });
+  }
+
+
+
+  public Command stop() {
```
markpete: Maybe rename to stopPivotMotor() and add a stopRollers() method?

markpete: Or add the stop for rollers in this method if that is desireable.

markpete (dismissed):

src/main/java/frc/robot/subsystems/Intake/IntakeArm.java
```diff
@@ -0,0 +1,139 @@
...
+            pivotSim = new ArmSim(pivotMotor);
+        }
+    }
+
+    // configs
+    private void TalonFXConfigs() {
+        // define config variables
+        var talonFXConfigs = new TalonFXConfiguration();
+        var simConfigs = talonFXConfigs.Slot0;
+        var irlConfigs = talonFXConfigs.Slot0;
```
Jetblackdragon: you are using the same slot you need to make them separate 
Ex. 
var simConfigs = new Slot0Configs();
var irlConfigs = new Slot0Configs();

src/main/java/frc/robot/subsystems/Intake/RollerSim.java
```diff
@@ -0,0 +1,75 @@
...
+  private final TalonFXSimState rightRollerSimState;
+
+  private final MechanismLigament2d roller;
+
+  private final double ROLLERS_GEAR_RATIO = 1.0; // tune this
+
+  // flywheel system variable
+  LinearSystem rollerSystem =
+      LinearSystemId.createFlywheelSystem(
+          DCMotor.getKrakenX60(2), 0.002, ROLLERS_GEAR_RATIO); // idk how to calculate MOI
```
Jetblackdragon: Might be wrong but this looks weird you define rollerSysten with 2 motors but then later on lines 41-42 you make 2 seperate roller sims with 1 motor each instead of 1 sim with 2 motors. Something to look at.

src/main/java/frc/robot/subsystems/Intake/IntakePivot.java
```diff
@@ -0,0 +1,146 @@
...
+  // Motion Magic Config
+  private static final double CRUISE_VELOCITY = 25;
+  private static final double ACCELERATION = 10;
+  private static final double JERK = 50;
+
+  // Gear Ratio
+  private static final double GEAR_RATIO = 36;
+
+  // Soft Limits
+  private static final double PIVOT_MAX = 180; // degrees
```
jamesdooley4: Likely should be just more than the deploy position (0.28). So, maybe 0.30

src/main/java/frc/robot/subsystems/Intake/IntakePivot.java
```diff
@@ -0,0 +1,146 @@
...
+
+    pivotMotor.getConfigurator().apply(config);
+  }
+
+  private void networktables() {
+    var nt = NetworkTableInstance.getDefault();
+    this.pivotTopic = nt.getDoubleTopic("intake/pivotPosition");
+    this.pivotPub = pivotTopic.publish();
+
+    pivotPub.set(0.0); // default value
```
jamesdooley4: Nothing publishes to this below. I'd recommend adding a 2nd topic for the pivot target position, and then create a periodic() function that sets pivotPub and the new target position topic to the current pivot position & the current target position.

src/main/java/frc/robot/Controls.java
```diff
@@ -221,7 +230,12 @@ private void configureLauncherBindings() {
                 s.hood.hoodPositionCommand(0.0), s.flywheels.setVelocityCommand(0.0)));
     driverController
         .y()
-        .onTrue(Commands.parallel(s.hood.zeroHoodCommand(), s.turretSubsystem.zeroTurret()));
+        .onTrue(
+            Commands.parallel(
+                    s.hood.zeroHoodCommand(),
+                    s.turretSubsystem.zeroTurret(),
+                    s.intakePivot.zeroPivot())
+                .ignoringDisable(true));
```
jamesdooley4: Please add a name for this command, so it's not just named "ParallelCommand"

jamesdooley4 (approved):
Just two minor comments left.

src/main/java/frc/robot/Controls.java
```diff
@@ -320,7 +351,5 @@ private void configureTurretBindings() {
         .onTrue(
             s.drivebaseSubsystem.runOnce(
                 () -> s.drivebaseSubsystem.resetPose(new Pose2d(13, 4, Rotation2d.kZero))));
-    driverController.povUp().whileTrue(s.turretSubsystem.rotateToHub());
-    driverController.povDown().onTrue(s.turretSubsystem.zeroTurret());
```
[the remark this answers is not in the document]
Jetblackdragon: part of the right trigger binding so no longer needed

RobototesProgrammers (approved):

PR #26, opened 2026-01-31
Moving shot

Nicole-N-Pedersen:
calculates time of shot from turret to the hub

### DIFF
--- a/src/main/java/frc/robot/subsystems/LauncherConstants.java
+++ b/src/main/java/frc/robot/subsystems/LauncherConstants.java
@@ -85,7 +85,7 @@ public class LauncherConstants {
 
   // Move a target a set time in the future along a velocity defined by fieldSpeeds
   public static Translation2d predictTargetPos(
-      Translation2d target, ChassisSpeeds fieldSpeeds, Double timeOfFlight) {
+      Translation2d target, ChassisSpeeds fieldSpeeds, double timeOfFlight) {
     double predictedX = target.getX() - fieldSpeeds.vxMetersPerSecond * timeOfFlight;
     double predictedY = target.getY() - fieldSpeeds.vyMetersPerSecond * timeOfFlight;
 


src/main/java/frc/robot/subsystems/LauncherConstants.java
```diff
@@ -6,71 +6,106 @@
...
-    public final double flywheelPower;
-    public final double distance;
-
-    public LauncherDistanceDataPoint(
-        double m_hoodAngle, double m_flywheelPower, double m_distance) {
-      this.hoodAngle = m_hoodAngle;
-      this.flywheelPower = m_flywheelPower;
-      this.distance = m_distance;
+  private static double DistanceVelocityArray[] = {1.0, 2.0, 3.0, 4.0, 5.0};
+    private static double MaxFieldDistance = DistanceVelocityArray.length-1;
```
jamesdooley4: Should remove these two since they're never used.

src/main/java/frc/robot/subsystems/LauncherConstants.java
```diff
@@ -6,71 +6,106 @@
...
-  private static final java.util.List<LauncherDistanceDataPoint> distanceData = new ArrayList<>();
+    public static double getFlywheelSpeedFromDistance(double distance) {
+      return flywheelMap.get(distance);
+    }
 
-  private static InterpolatingDoubleTreeMap flywheelMap = new InterpolatingDoubleTreeMap();
-  private static InterpolatingDoubleTreeMap hoodMap = new InterpolatingDoubleTreeMap();
+    public static double getFlywheelSpeedFromPose2d(Translation2d hub, Pose2d robot) {
+      Transform2d fieldRelativeLauncherOffset = new Transform2d(LAUNCHER_OFFSET, robot.getRotation());
+      robot = robot.plus(fieldRelativeLauncherOffset);
```
jamesdooley4: This is repeated a few times, so would be best to factor into its own method like `public static Translation2d launcherFromRobot(Pose2d robot)`

src/main/java/frc/robot/subsystems/LauncherConstants.java
```diff
@@ -6,71 +6,106 @@
...
+          double predictedX = target.getX() - fieldSpeeds.vxMetersPerSecond * timeOfFlight;
+          double predictedY = target.getY() - fieldSpeeds.vyMetersPerSecond * timeOfFlight;
+
+          return new Translation2d(predictedX, predictedY);
+      }
+
+      public static Translation2d iterativeMovingShotFromFunnelClearance(
+              Pose2d robot, ChassisSpeeds fieldSpeeds, Translation2d target, int iterations) {
+          // Perform initial estimation (assuming unmoving robot) to get time of flight estimate
+          double distance = robot.getTranslation().getDistance(target);
```
jamesdooley4: Here & below need to account for the offset of the launcher. With the new `launcherFromRobot()` method (earlier suggestion), that'd just be:
```java
double distance = launcherFromRobot(robot).getDistance(target);
```

PR #28, opened 2026-01-31
Done some more fixes/additions for LED_lights.java

SnappyRiffs:
Need to chnge flickering and keybind stuff. shouldnt be too hard as i have a method to turn off the leds

### DIFF
--- a/src/main/java/frc/robot/Controls.java
+++ b/src/main/java/frc/robot/Controls.java
@@ -258,7 +258,7 @@ public class Controls {
           () -> {
             mode = LEDMode.OUTTAKE_IN_PROGRESS;
           });
-    );
+    )
     
     driveController
     .leftTrigger()
@@ -267,7 +267,7 @@ public class Controls {
           () -> {
             mode = LEDMode.INTAKE_IN_PROGRESS;
           });
-    );
+    )
     // TODO: Add button to set climb and set default
     
     if (mode == LedSubsystem.LEDMode.INTAKE_IN_PROGRESS) { // blue


src/main/java/frc/robot/sensors/LED_lights.java
```diff
@@ -0,0 +1,154 @@
...
+import edu.wpi.first.units.measure.Frequency;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.Commands;
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+import edu.wpi.first.wpilibj2.command.RunCommand;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.util.LEDMode;
+import java.util.function.Supplier;
+
+public class LED_lights extends SubsystemBase {
```
jamesdooley4: Since this is a subsystem, should include that in the name like the other subsystems. Maybe `LEDSubsystem`?

jamesdooley4: I'd also recommend moving this to the 'subsystems' folder since it does output like subsystems with actuators, vs. sensors that only provide inputs.

src/main/java/frc/robot/Controls.java
```diff
@@ -182,6 +190,36 @@ private void configureAutoAlignBindings() {
     driverController.rightBumper().whileTrue(FuelAutoAlign.autoAlign(this, s));
   }
 
+  private void configureLEDBindings() {
+    LED_lights LEDs = s.led_Lights;
```
jamesdooley4: Needs to detect the LED subsystem being null (if it's disabled) and then return early to avoid crashes below.

src/main/java/frc/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Sets Leds at specified {@link frc.robot.subsystems.LEDSubsystem#CAN_ID id} to color
+   *
+   * @param color
+   */
+  public Command updateLEDs(RGBWColor color) {
+    return Commands.run(
+            () -> {
+              SolidColor solid = new SolidColor(0, END_INDEX);
```
markpete: I recommend making this a private class object instead of reallocating memory for it on every updateLEDs() call as it generates a lot of memory garbage.

src/main/java/frc/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,177 @@
...
+              solid.withColor(color);
+              candle.setControl(solid);
+            })
+        .withName("Set leds to specified color");
+  }
+
+  /* to be used in the future */
+  public Command LED_Mode(Supplier<LEDMode> mode) {
+    return Commands.run(
```
markpete: Command Scheduler Flooding
The code in this routine schedules a command inside a Commands.run loop.

Problem: Commands.run executes its body every loop cycle (20ms). This code calls CommandScheduler.getInstance().schedule(...) ~50 times per second.
Consequence: This will flood the scheduler with new commands. Since updateLEDs creates a RunCommand (which never finishes), the scheduler will be overwhelmed handling interruptions and new requirements, likely causing the robot code to crash or become unresponsive.

src/main/java/frc/robot/subsystems/LEDSubsystem.java
SnappyRiffs: I use the updateLEDs in robot.java for setting the default color (red when teleop starts)

src/main/java/frc/robot/subsystems/LEDSubsystem.java
SnappyRiffs: this code will be used later down the line after i test everything with frc driverstation and fix some funky things happening with the leds

SnappyRiffs:
will fix formatting once i am back at the meeting since gradlew fails on my personal computer for whatever reason

src/main/java/frc/robot/Controls.java
```diff
@@ -34,11 +35,15 @@ public class Controls {
   private static final int FEEDER_TEST_CONTROLLER_PORT = 1;
   private static final int SPINDEXER_TEST_CONTROLLER_PORT = 2;
   private static final int LAUNCHER_TUNING_CONTROLLER_PORT = 3;
+  private static final int LED_CONTROLLER_PORT = 4;
```
iamawesomecat: will conflict w other subsystems. Either have it set to like 5 or something or something else.

src/main/java/frc/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,179 @@
...
+  public final Command rainbowCommand;
+
+  private RGBWColor activeSolidColor = OFF_COLOR;
+
+  public LEDSubsystem() {
+    setRainbowAnimation(0, 1, AnimationDirectionValue.Forward, Units.Hertz.of(100));
+    // Create persistent rainbow command
+    rainbowCommand =
+        Commands.run(() -> candle.setControl(slot0Animation), this)
+            .repeatedly()
```
markpete: Redundant Command Composition (Performance/Best Practice)
Description: The rainbowCommand is defined as Commands.run(...).repeatedly().
Commands.run(...) creates a RunCommand, which by definition runs endlessly until interrupted.
Adding .repeatedly() wraps it in a RepeatedCommand, which effectively does nothing for a RunCommand but adds a wrapper layer.
Recommendation: Remove .repeatedly(). Commands.run(..., this) is sufficient.

(Test this out to confirm I'm correct...  I may be wrong on this.)

src/main/java/frc/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,150 @@
...
+   */
+  public void setRainbowAnimation(
+      int slot, double brightness, AnimationDirectionValue direction, Frequency frameRate) {
+
+    slot0Animation
+        .withSlot(slot)
+        .withBrightness(brightness)
+        .withDirection(direction)
+        .withFrameRate(frameRate);
+  }
```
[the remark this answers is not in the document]
SnappyRiffs: will rename variable

src/main/java/frc/robot/Controls.java
```diff
@@ -199,6 +216,42 @@ private void configureAutoAlignBindings() {
...
+    LEDSubsystem LEDs = s.ledSubsystem;
+    if (s.ledSubsystem == null) {
+      // Stop running this method
+      return;
+    }
+    ledTestController
+        .a()
+        .onTrue(
+            LEDs.setLEDsCommand(LEDSubsystem.OFF_COLOR)
+                .andThen(LEDs.setLEDsCommand(LEDSubsystem.INTAKE_COLOR))); // intake color yellow
```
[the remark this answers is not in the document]
SnappyRiffs: i think it is better with the `LEDs.setLEDsCommand(LEDSubsystem.OFF_COLOR)` imo

iamawesomecat:
have you considered resolving comments

src/main/java/frc/robot/Controls.java
```diff
@@ -201,6 +218,64 @@ private void configureAutoAlignBindings() {
...
+      // Stop running this method
+      return;
+    }
+    ledTestController
+        .a()
+        .onTrue(
+            LEDs.setLEDsCommand(LEDSubsystem.OFF_COLOR, LEDSubsystem.GLOBAL_BRIGHTNESS)
+                .andThen(
+                    LEDs.setLEDsCommand(LEDSubsystem.INTAKE_COLOR, LEDSubsystem.GLOBAL_BRIGHTNESS))
+                .withName("Set intake color")); // blue
```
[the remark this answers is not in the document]
SnappyRiffs: the flickering issue is no more gemini!

src/main/java/frc/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,359 @@
...
+   * }</pre>
+   *
+   * @param color the original {@link RGBWColor} to scale
+   * @param brightness a value ideally between {@code 0.0} and {@code 1.0}
+   * @return a new {@link RGBWColor} with each channel scaled by the clamped brightness factor
+   */
+  private RGBWColor scaleBrightness(RGBWColor color, double brightness) {
+    brightness = Math.max(0.0, Math.min(1.0, brightness)); // clamp 0–1
+
+    return new RGBWColor(
```
markpete: Is there a reason to do it this way and create a new color instead of just using .withBrightness(newBrightness)?

src/main/java/frc/robot/subsystems/LEDSubsystem.java
```diff
@@ -0,0 +1,359 @@
...
+  // private final StringTopic alternatingColors;
+
+  /**
+   * Publisher for {@link #alternatingColors}, used to send the current alternating colors to
+   * dashboards.
+   */
+  // private final StringPublisher alternatingColorsPub;
+
+  /** */
+  public void LEDSubsystem() {
```
markpete: Constructor Logic Error (Mismatched Name & Void Return Type)

Description: The intended constructor has a mismatched name (LEDSubsystem vs the class name LedSubsystem) and includes a void return type. In Java, this makes it a regular method, not a constructor. As a result, this method is never called when the subsystem is instantiated, and initialization code such as setRainbowAnimation(...) is skipped.
Recommendation: Remove void and rename the method to match the class name exactly: public LedSubsystem() { ... }

markpete (changes requested):
Lots of commented out code in this PR, so it's a little hard to follow.  If the code is still in testing and not ready for code reviews, please mark it as "Draft", so that I know it's not ready for me to look at.

SnappyRiffs:
I have it so that at auto it is rainbowanimation

default color (red): left bumper
climb (cyan): right bumper
intake (blue): left trigger
shoot (green): right trigger

i will merge main

PR #44, opened 2026-02-09
Issue 42 fix

### LINKED ISSUE #42: updateSim value of 20ms should be moved to a common location
Currently many of the *Sim.java files set an update value of 20ms locally.  We should move this to a central location and have them all read from it.  This way we can update in a single place instead of having to update every Sim file.

iamawesomecat:
Fixes #42


markpete (approved):
A few suggestions for improvements, but these were all in the prior code, so consider these [nit] wish-list items.  Approving with these comments.

PR #56, opened 2026-02-14
Feeder subsystem motor control

kyladev:


src/main/java/frc/robot/subsystems/FeederSubsystem.java
```diff
@@ -17,16 +20,9 @@
...
   public static final int feederRumbleThreshold = 67;
-
-  private static final double FEEDMOTOR_KP = 38.5;
-  private static final double FEEDMOTOR_KI = 0;
-  private static final double FEEDMOTOR_KD = 0;
-  private static final double FEEDMOTOR_KS = 0;
-  private static final double FEEDMOTOR_KV = 0;
-  private static final double FEEDMOTOR_KG = 0;
-  private static final double FEEDMOTOR_KA = 0;
+  public TorqueCurrentFOC torqueReq = new TorqueCurrentFOC(0);
```
markpete: Recommend this be made a private instead of public.  This shouldn't be exposed outside of this class.

src/main/java/frc/robot/subsystems/FeederSubsystem.java
```diff
@@ -17,16 +20,9 @@
 
 public class FeederSubsystem extends SubsystemBase {
   private int ballsDetectedNum = 0;
-  public static final double feederSpeed = 0.4;
+  public static final double feederCurrent = 12;
```
markpete: [nit] Constants should be in CAPS to match standard naming.  Recommend you rename feederCurrent to FEEDER_CURRENT

src/main/java/frc/robot/subsystems/index/SpindexerSubsystem.java
```diff
@@ -17,25 +17,27 @@
...
           LinearSystemId.createFlywheelSystem(
               DCMotor.getKrakenX60(1),
               0.001,
               1.0); // TODO: Update to final moment of inertia and gear ratio
       motorSim =
           new FlywheelSim(
-              serialMotorSystem, DCMotor.getKrakenX60(1), 1.0); // TODO: Update to final gear ratio
+              spindexerMotorSystem,
+              DCMotor.getKrakenX60(1),
+              1.0); // TODO: Update to final gear ratio
```
markpete: Do we know what the gear ratio will be yet based on the CAD?

src/main/java/frc/robot/subsystems/index/SpindexerSubsystem.java
```diff
@@ -62,31 +64,25 @@ public void spindexerConfig() {
   }
 
   public void setSpeed(double speed) {
```
markpete: Passing a raw percentage back via spindexerMotor.set(speed) natively constructs a new internal DutyCycleOut control request internally on every loop iteration as per the Phoenix 6 API changes. Since this routine is meant to be called in an execute loop at 50Hz, it will incur consistent garbage collection spikes over long durations. (It is notable that this issue was properly mitigated in FeederSubsystem in this PR).

Recommendation: Define, cache, and reuse the control request within SpindexerSubsystem similar to what was implemented for FeederSubsystem's voltage equivalent.

private final DutyCycleOut dutyCycleReq = new DutyCycleOut(0);

public void setSpeed(double speed) {
  spindexerMotor.setControl(dutyCycleReq.withOutput(speed));
}

src/main/java/frc/robot/subsystems/index/IndexerSubsystem.java
```diff
@@ -15,12 +15,12 @@ public IndexerSubsystem(FeederSubsystem feeder, SpindexerSubsystem spindexerSubs
   public Command runIndexer() {
     return Commands.runEnd(
         () -> {
-          feeder.startMotor();
-          spindexerSubsystem.startMotor();
+          feeder.setVoltage(FeederSubsystem.FEEDER_VOLTAGE);
+          spindexerSubsystem.setSpeed(SpindexerSubsystem.SPINDEXER_SPEED);
         },
         () -> {
```
markpete: In WPILib, a Command object is essentially a set of instructions that the CommandScheduler will run later. Calling a method like stopMotorCommand() creates those instructions, but it doesn't actually execute them right away.

Let's look closely at the runIndexer command factory in IndexerSubsystem.java from the PR:
```
public Command runIndexer() {
    return Commands.runEnd(
        // 'run' Runnable
        () -> {
          feeder.setVoltage(FeederSubsystem.FEEDER_VOLTAGE);
          spindexerSubsystem.setSpeed(SpindexerSubsystem.SPINDEXER_SPEED);
        },
        // 'end' Runnable
        () -> {
          feeder.stopMotorCommand(); // <-- Creates a Command, but doesn't run it!
          spindexerSubsystem.stopMotorCommand(); // <-- Creates a Command, but doesn't run it!
        });
  }
```
Commands.runEnd expects two Runnable actions. Inside the second Runnable (the end action), you are calling feeder.stopMotorCommand().

If we look at FeederSubsystem:
```
public Command stopMotorCommand() {
    return runOnce(() -> feedMotor.stopMotor());
  }
```
This method returns a Command object. However, because you are calling it inside the runEnd callback, the created Command object is simply returned to the void and discarded. It is never passed to WPILib's CommandScheduler to be executed. As a result, feedMotor.stopMotor() is never actually called, and the motors will keep spinning forever when runIndexer ends.

The Fix
Instead of calling command factories inside your runEnd, you need to perform the direct motor actions.

1. Create a method in your subsystems that actually stops the motor immediately (returns void):
// In FeederSubsystem.java
```
public void stopMotor() {
    feedMotor.stopMotor();
}
```
// In SpindexerSubsystem.java
```
public void stopMotor() {
    spindexerMotor.stopMotor();
}
```
2. Call those direct action methods in your runEnd instead:
// In IndexerSubsystem.java
```
  public Command runIndexer() {
    return Commands.runEnd(
        () -> {
          feeder.setVoltage(FeederSubsystem.FEEDER_VOLTAGE);
          spindexerSubsystem.setSpeed(SpindexerSubsystem.SPINDEXER_SPEED);
        },
        () -> {
          feeder.stopMotor(); // <-- This directly stops the motor now
          spindexerSubsystem.stopMotor(); // <-- This directly stops the motor now
        });
  }
```
This ensures that when the command ends, the motors are actually commanded to stop right then and there. Let me know if that makes sense!

markpete (changes requested):

Issue #43, opened 2026-02-08
Replace System.Out for debug logging


markpete:
Using System.Out for debugging and logging seems to be a common practice within the team.  System.Out is actually very resource intensive, especially in loops or commonly called routines.  We should explore a library or common best practice method for our team to achieve this, but in a lower cost way.

Consider:  https://training.teamresistance.org/telemetry/  Or similar.

PR #59, opened 2026-02-15
[Style] Blocking Print Commands

markpete:
System.Out.printX() can cause performance issues for a robot at run-time and is 300x worse than without it in code.  In general, it's better to use DataLogManager.log().  (See Issue #43 for details)

By default - code will fail to compile and recommend the developer to use DataLogManager.log() instead of System.out.println().  However, this can be bypassed or overridden with a comment of //print-allow-override for cases where it is truly needed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,7 @@ plugins {
     id "edu.wpi.first.GradleRIO" version "2026.2.1"
     id "com.diffplug.spotless" version "6.25.0"
     id("io.freefair.lombok") version "9.2.0"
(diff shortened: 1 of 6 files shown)


jamesdooley4:
FYI, DataLogManager.log() internally calls `System.out.println`. See https://github.com/wpilibsuite/allwpilib/blob/77dfad97c66fd4922fb19fd7f805c96357ea13cb/wpilibj/src/main/java/edu/wpi/first/wpilibj/DataLogManager.java#L170

PR #60, opened 2026-02-17
Super Structure implementation

Jetblackdragon:


src/main/java/frc/robot/subsystems/launcher/LauncherSubsystem.java
```diff
@@ -0,0 +1,74 @@
...
+          hoodGoal = LauncherConstants.getHoodAngleFromPose2d(targetPose, drive.getState().Pose);
+          flywheelsGoal =
+              LauncherConstants.getFlywheelSpeedFromPose2d(targetPose, drive.getState().Pose);
+
+          hoodGoalPub.set(hoodGoal);
+          flywheelGoalPub.set(flywheelsGoal);
+
+          hood.setHoodPosition(hoodGoal);
+          flywheels.setVelocityRPS(flywheelsGoal);
+          CommandScheduler.getInstance().schedule(turret.rotateToHub());
```
jamesdooley4: This will schedule a new command 50x a second since the first parameter to `runEnd` is the code that runs every loop. This should instead just call the turret subsystem to tell it what field-relative angle to point at (equivalent of the hood & flywheel calls above to set their goal position or RPS).

jamesdooley4 (approved):
Looks good.

PR #73, opened 2026-02-27
update led pr

Jetblackdragon:


src/main/java/frc/robot/sensors/LEDSubsystem.java
```diff
@@ -0,0 +1,316 @@
...
+   * active animation is cleaRed.
+   *
+   * @param enabled true to start the rainbow animation, false to stop it
+   */
+  public void setRainbowEnabled(boolean enabled) {
+    if (enabled) {
+      candle.setControl(rainbowAnimation);
+      isRainbowPub.set(true);
+      currentAnimationPub.set("Rainbow");
+      currentColorPub.set("None"); // always set to None
```
markpete: Move this line outside of the if/else and then just call it in either case.  It simplifies your if() and reduces lines of code.

src/main/java/frc/robot/Controls.java
```diff
@@ -347,4 +351,11 @@ private void configureTurretBindings() {
             s.turretSubsystem.pointFacingJoystick(
                 () -> driverController.getRightX(), () -> driverController.getRightY()));
   }
+
+  public void configureLedBindings() {
+    if (s.ledSubsystem == null) {
+      return;
+    }
+    s.ledSubsystem.setDefaultCommand(s.ledSubsystem.setLEDsCommand(LEDSubsystem.OFF_COLOR));
```
markpete: CommandScheduler Thrashing with LED Default Command
Severity: High (Performance & Logic Concern)

`s.ledSubsystem.setDefaultCommand(s.ledSubsystem.setLEDsCommand(LEDSubsystem.OFF_COLOR));`

Description: LEDSubsystem.setLEDsCommand(...) returns a command using Commands.runOnce(...). Commands created with runOnce finish immediately after executing. When an instant-finishing command is set as a subsystem's Default Command, the CommandScheduler constantly sees that the command has finished and re-initializes/re-schedules it every single 20ms robot loop. This leads to thrashing the garbage collector as it continuously creates and destroys command execution instances, while simultaneously spamming the CAN bus with updates to the CANdle every 20ms.

Suggestion: Avoid using runOnce inside a default command. If the goal is to continuously ensure the LEDs are off when no other active command is using them, use a continuous command: `Commands.run(() -> s.ledSubsystem.setHardwareColor(LEDSubsystem.OFF_COLOR), s.ledSubsystem). `However, an even better pattern for LEDs is to only send the command over the CAN bus when the color state actually changes rather than constantly forcing it.

src/main/java/frc/robot/sensors/LEDSubsystem.java
```diff
@@ -0,0 +1,316 @@
...
+   * Sets the CANdle LED controller to a solid color at the specified brightness.
+   *
+   * <p>This method directly updates the physical hardware output. It is intended to be called
+   * internally by higher-level commands that manage LED behavior.
+   *
+   * @param color the {@link RGBWColor} to apply to the LED strip
+   * @param brightness the brightness scalar (0.0–1.0) applied to the color
+   */
+  public void setHardwareColor(RGBWColor color, double brightness) {
+    RGBWColor scaled = scaleBrightness(color, brightness);
```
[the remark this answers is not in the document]
Roxie222: looking at docs, withBrightness is reserved for animations but not for rgbwcolor

PR #76, opened 2026-03-01
Clock and hub status

### LINKED ISSUE #106: Put LED rainbow back once it plays nice with other LED configurations
> Need to figure out issues with rainbow and put it back (or not). 

 _Originally posted by @jamesdooley4 in [#64](https://github.com/robototes/REBUILT2026/pull/64/changes#r2928778752)_

### LINKED ISSUE #184: DriveState pose can go outside field walls
If the robot is pushed near a field wall or the robot drives into a field wall and spins its wheels, the robot's pose will leave the field entirely. This issue tracks doing <something> to keep the robot's pose within the field so wildly incorrect poses aren't being used for the launch calculator, or causing Vision to discard data because the vision poses are "too far off" from where the drivestate thinks the robot is.

### LINKED ISSUE #202: HubShiftUtil.getShiftedShiftInfo should use precached shiftedShiftStartTimes, shiftedShiftEndTimes arrays
To avoid a bunch of computation and creating (and destroying) two arrays multiple times each loop, HubShiftUtil.getShiftedShiftInfo should pre-cache shiftedShiftStartTimes and shiftedShiftEndTimes arrays for both the 'red wins auto' and 'blue wins auto' cases, and then pick between them at runtime and pass them in `getShiftInfo()`.

Roxie222:


src/main/java/frc/robot/util/ClockAndHubStatus.java
```diff
@@ -0,0 +1,119 @@
...
+    switch (gameData.charAt(0)) {
+      case 'R' -> redInactiveFirst = true;
+      case 'B' -> redInactiveFirst = false;
+      default -> {
+        // If we have invalid game data, assume hub is active.
+        return true;
+      }
+    }
+
+    // Shift was is active for blue if red won auto, or red if blue won auto.
```
markpete: // Shift is active

markpete (changes requested):

src/main/java/frc/robot/util/HubShiftUtil.java
```diff
@@ -0,0 +1,189 @@
...
+        10.0,
+        35.0 + endingActiveFudge,
+        60.0 + approachingActiveFudge,
+        85.0 + endingActiveFudge,
+        110.0 + approachingActiveFudge,
+        140.0
+      };
+      return getShiftInfo(shiftSchedule, shiftedShiftStartTimes, shiftedShiftEndTimes);
+    }
+    double[] shiftedShiftStartTimes = {
```
markpete: Move these two arrays to finals too outside of the routine.

src/main/java/frc/robot/util/HubShiftUtil.java
```diff
@@ -0,0 +1,189 @@
...
+        return Alliance.Red;
+      }
+    }
+
+    // Return default value
+    return alliance == Alliance.Blue ? Alliance.Red : Alliance.Blue;
+  }
+
+  /** Starts the timer at the begining of teleop. */
+  public static void initialize() {
```
markpete: I don't believe anything is calling initialize(), so the timer will never actually be started.

markpete (changes requested):

src/main/java/frc/robot/Robot.java
```diff
@@ -226,11 +229,16 @@ public void teleopInit() {
     // teleop starts running. If you want the autonomous to
     // continue until interrupted by another command, remove
     // this line or comment it out.
+<<<<<<< HEAD
+    HubShiftUtil instance = new HubShiftUtil();
+    instance.initialize();
+=======
```
markpete: Merge conflict

markpete (changes requested):

src/main/java/frc/robot/sensors/LEDSubsystem.java
```diff
@@ -117,7 +117,7 @@ private void setSolid(RGBWColor color) {
     primaryColor = color;
     setHardwareColor(color);
 
-    currentPatternPub.set("SOLID");
+    currentPatternPub.set(currentPattern.name());
```
markpete: [nit] I don't think you need the currentPattern variable.  You can just do:

`currentPatternPub.set(LEDPattern.SOLID.name());`

src/main/java/frc/robot/util/ClockAndHubStatus.java
```diff
@@ -0,0 +1,187 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import frc.robot.subsystems.drivebase.CommandSwerveDrivetrain;
+import java.util.Optional;
+import java.util.function.Supplier;
+import lombok.Setter;
+
+public class ClockAndHubStatus {
+  public double matchLength = 2.5;
+
+  private Translation2d pointLeftFieldTop = new Translation2d(2, 6);
```
markpete: Should the field variables be in a Field.java file?  Will other subsystems want or need this info?

src/main/java/frc/robot/util/ClockAndHubStatus.java
```diff
@@ -0,0 +1,187 @@
...
+        return pointRightFieldTop;
+      } else {
+        return pointRightFieldBottom;
+      }
+    } else {
+      return AllianceUtils.getHubTranslation2d();
+    }
+  }
+
+  /** Starts the timer at the begining of teleop. */
```
markpete: [nit] spelling:  "beginning"

markpete (dismissed):

src/main/java/frc/robot/sensors/LEDSubsystem.java
```diff
@@ -132,22 +153,24 @@ private void setAlternating(RGBWColor a, RGBWColor b, double interval) {
     showingPrimary = true;
     lastToggleTime = Timer.getFPGATimestamp();
 
-    currentPatternPub.set("ALTERNATING");
-    currentColorPub.set(a.toHexString());
+    currentPatternPub.set(currentPattern.name());
+    currentColorPub.set("NONE");
     alternatingColorsPub.set("A:" + a.toHexString() + " | B:" + b.toHexString());
```
jamesdooley4: These duplicate the `publishState()` call below. Pick one to keep.

src/main/java/frc/robot/sensors/LEDSubsystem.java
```diff
@@ -132,22 +153,24 @@ private void setAlternating(RGBWColor a, RGBWColor b, double interval) {
...
   }
 
   private void setRainbow() {
     currentPattern = LEDPattern.RAINBOW;
 
     // Update publishers
-    currentPatternPub.set("RAINBOW");
+    currentPatternPub.set(currentPattern.name());
     currentColorPub.set("NONE");
     alternatingColorsPub.set("A:None | B:None");
```
jamesdooley4: These duplicate the `publishState()` call below. Pick one to keep.

src/main/java/frc/robot/sensors/LEDSubsystem.java
```diff
@@ -113,16 +132,18 @@ public Command setLEDsCommand(RGBWColor color) {
...
   private void setSolid(RGBWColor color) {
+    candle.setControl(emptyAnimation);
     currentPattern = LEDPattern.SOLID;
     primaryColor = color;
+
     setHardwareColor(color);
 
-    currentPatternPub.set("SOLID");
     currentColorPub.set(color.toHexString());
     alternatingColorsPub.set("A:None | B:None");
```
jamesdooley4: Either put back the `currentPatternPub.set()` call, or replace these lines with `publishState()`

src/main/java/frc/robot/subsystems/launcher/LauncherSubsystem.java
```diff
@@ -52,7 +53,9 @@ public boolean isAtTarget() {
         && s.hood.atTargetPosition()
         && s.turretSubsystem.atTarget()
         && !LaunchCalculator.isApproachingTrench(
-            s.drivebaseSubsystem.getState().Pose, s.drivebaseSubsystem.getState().Speeds);
+            s.drivebaseSubsystem.getState().Pose, s.drivebaseSubsystem.getState().Speeds)
+        && !LaunchCalculator.isUnderClimb(
+            s.drivebaseSubsystem.getState().Pose.transformBy(LauncherConstants.turretTransform()));
```
jamesdooley4: This now calls `s.drivebaseSubsystem.getState()` three times in rapid succession if everything is ready to launch and the robot isn't in a dangerous spot. I'd recommend calling `s.drivebaseSubsystem.getState()` once and caching the result to use in these three places.

src/main/java/frc/robot/subsystems/VisionSubsystem.java
```diff
@@ -263,8 +263,13 @@ private void processLimelight(
 
     if (estimate == null || estimate.tagCount <= 0) return;
 
-    rawFieldPoseEntry.set(estimate.pose3d);
     Pose2d visionPose2d = estimate.pose3d.toPose2d();
+    if (estimate.timestampSeconds == camera.getLastTimestampSeconds()) {
+      publishDiagnostics(estimate, visionPose2d, camera, "stale-timestamp");
+      return;
+    }
+
```
jamesdooley4, line 271: Due to the latency value jumping around every cycle inside `getBetterBotPoseEstimate`, the `estimate.timestampSeconds` value is never exactly equal to the last timestamp, so this code never hits the equivalence case. So, I'd recommend deleting it.

src/main/java/frc/robot/util/LLCamera.java
```diff
@@ -133,6 +135,15 @@ private static BetterPoseEstimate getBetterBotPoseEstimate(
       // Handle the case where no data is available
       return null; // or some default PoseEstimate
     }
+    double[] lastRaw = isMegaTag2 ? lastRawBotposeMT2 : lastRawBotposeMT1;
+    if (lastRaw != null && java.util.Arrays.equals(poseArray, lastRaw)) {
```
jamesdooley4, line 139: This doesn't actually work, since entry 6 (latency) keeps changing every loop even if everything else is identical. It's probably effective enough to just compare the fix six entries that make up the pose and early-exit if they are identical since it's nearly impossible for the vision solver to come up with the same answer due to inherent noise in the camera frames.

src/main/java/frc/robot/Controls.java
```diff
@@ -242,10 +252,23 @@ private void configureLauncherBindings() {
 
     driverController
         .rightTrigger()
+        .or(GetTargetFromPose.autoShoot(s.drivebaseSubsystem))
```
jamesdooley4: This trigger should be created once as a class variable in `Controls` and the referenced here and down in `configureLedBindings()` so it's only evaluated once per loop rather than twice.

src/main/java/frc/robot/Controls.java
```diff
@@ -242,10 +252,23 @@ private void configureLauncherBindings() {
...
+        .and(new Trigger(() -> !driverController.a().getAsBoolean()))
```
jamesdooley4: This is reinventing existing functionality. This should be equivalent and be cleaner:
```suggestion
        .and(driverController.a().negate())
```

src/main/java/frc/robot/Controls.java
```diff
@@ -426,6 +459,16 @@ private void configureTurretBindings() {
...
+    driverController
+        .y()
+        .toggleOnTrue(
+            s.turretSubsystem
+                .run(
+                    () ->
+                        s.turretSubsystem.setTurretRawPosition(
+                            s.turretSubsystem.getTurretPosition()))
+                .withName("Turret Kill — Hold Position"))
+        .onTrue(Commands.runOnce(() -> turretKillActive = !turretKillActive));
```
jamesdooley4: This command needs a name too.

src/main/java/frc/robot/Controls.java
```diff
@@ -473,19 +516,16 @@ public void configureLedBindings() {
...
-            () -> {
-              if (RobotState.isAutonomous()) return false;
-
-              var shiftInfo = HubShiftUtil.getOfficialShiftInfo();
-
-              return shiftInfo.remainingTime() <= 4.0;
-            });
-
-    shiftWarning.onTrue(s.ledSubsystem.flashCommand(LEDSubsystem.CLIMB_COLOR, 5, 0.1));
+    Trigger leaveZoneWarning = new Trigger(GetTargetFromPose.autoShoot(s.drivebaseSubsystem));
```
jamesdooley4: This is creating a Trigger from a Trigger. Instead, just re-use the new class instance variable suggested to be added above in the comment in `configureLauncherBindings`.

src/main/java/frc/robot/Controls.java
```diff
@@ -292,6 +315,16 @@ private void configureLauncherBindings() {
...
 
+    driverController
+        .x()
+        .onTrue(
+            Commands.runOnce(() -> HubShiftUtil.setAllianceWinOverride(() -> Optional.of(false))));
+
+    driverController
+        .b()
+        .onTrue(
+            Commands.runOnce(() -> HubShiftUtil.setAllianceWinOverride(() -> Optional.of(true))));
```
jamesdooley4: These commands need names

src/main/java/frc/robot/Controls.java
jamesdooley4: It also needs to handle the case where `s.drivebaseSubsystem` is null. I'd recommend just having the Trigger always return `false` in that case.

src/main/java/frc/robot/Controls.java
```diff
@@ -426,6 +459,16 @@ private void configureTurretBindings() {
...
     s.turretSubsystem.setDefaultCommand(s.turretSubsystem.rotateToTargetWithCalc());
+    driverController
+        .y()
+        .toggleOnTrue(
+            s.turretSubsystem
+                .run(
+                    () ->
+                        s.turretSubsystem.setTurretRawPosition(
+                            s.turretSubsystem.getTurretPosition()))
+                .withName("Turret Kill — Hold Position"))
```
jamesdooley4, line 479: If this is interrupted (say, robot becoming disabled due to comms drop or radio reboot), then `turretKillActive` will be stuck on until 'y' is pressed again. If that's not intended, consider resetting that in disabledInit().

jamesdooley4 (changes requested):

src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
```diff
@@ -287,4 +295,21 @@ public double[] getWheelRotations() {
...
+    if (clampedX != current.getX() || clampedY != current.getY()) {
+      resetPose(new Pose2d(new Translation2d(clampedX, clampedY), current.getRotation()));
+    }
+  }
+
+  public boolean isBeached(double pitchThreshold) {
+    double pitch = Math.abs(getPigeon2().getPitch().getValueAsDouble());
+    double roll = Math.abs(getPigeon2().getRoll().getValueAsDouble());
+    return pitch > pitchThreshold || roll > pitchThreshold;
+  }
```
jamesdooley4, line 314: FYI that this only works in the 'beached' scenario where the robot is half-on fuel. In the 'surfing' scenario where it's riding on top of fuel, I expect the pitch & roll will be near zero.

src/main/java/frc/robot/subsystems/launcher/LauncherSubsystem.java
```diff
@@ -48,11 +50,13 @@ public boolean isAtTarget() {
...
     }
+    SwerveDriveState driveState = s.drivebaseSubsystem.getState();
     return s.flywheels.atTargetVelocity(flywheelsGoal, s.flywheels.FLYWHEEL_TOLERANCE)
         && s.hood.atTargetPosition()
         && s.turretSubsystem.atTarget()
-        && !LaunchCalculator.isApproachingTrench(
-            s.drivebaseSubsystem.getState().Pose, s.drivebaseSubsystem.getState().Speeds);
+        && !LaunchCalculator.isApproachingTrench(driveState.Pose, driveState.Speeds)
+        && !LaunchCalculator.isUnderClimb(
+            driveState.Pose.transformBy(LauncherConstants.turretTransform()));
```
jamesdooley4, line 59: It'd be better to do this internally in `isUnderClimb` to be consistent with `isApproachingTrench`.

src/main/java/frc/robot/Controls.java
```diff
@@ -478,19 +525,16 @@ public void configureLedBindings() {
...
-            () -> {
-              if (RobotState.isAutonomous()) return false;
-
-              var shiftInfo = HubShiftUtil.getOfficialShiftInfo();
-
-              return shiftInfo.remainingTime() <= 4.0;
-            });
-
-    shiftWarning.onTrue(s.ledSubsystem.flashCommand(LEDSubsystem.CLIMB_COLOR, 5, 0.1));
+    readyToShoot = GetTargetFromPose.autoShoot(s.drivebaseSubsystem);
```
jamesdooley4: Seems odd to set this only when the LED subsystem exists, but this does work since both the LED & drivebase subsystem are enabled.

jamesdooley4 (approved):
Still need to fix the issue with vision sometimes throttling during teleop & auto, but that's a pre-existing issue. Current changes look good - comments are just suggestions to make it even better.

src/main/java/frc/robot/Controls.java
```diff
@@ -247,10 +254,24 @@ private void configureLauncherBindings() {
 
     driverController
         .rightTrigger()
+        .or(readyToShoot)
```
jamesdooley4, line 259: This likely captures the 'always false' instance of `readyToShoot` since configureLauncherBindings happens before configureLedBindings where it's updated to the actual implementation

src/main/java/frc/robot/subsystems/auto/AutoDriveRotate.java
```diff
@@ -56,7 +64,6 @@ public AutoRotateCommand(
       addRequirements(drive);
     }
 
-    // TODO: Add auto rotate for launching game pieces to corners + climb alignment
     @Override
     public void initialize() {
       targetTranslation = AllianceUtils.getHubTranslation2d();
```
jamesdooley4, line 69: When this command is used, it only ever targets the hub. So, in conjuction with autoshoot, it'd aim at the hub even from the neutral zone if the shift is inactive.

jamesdooley4 (changes requested):
Missed that auto-shoot no longer works due to initialization ordering problem.

src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
RobototesProgrammers: any ideas on what to do in those scenarios?

src/main/java/frc/robot/subsystems/launcher/LauncherSubsystem.java
RobototesProgrammers: post dcmps optimization

src/main/java/frc/robot/subsystems/auto/AutoDriveRotate.java
RobototesProgrammers: fixed in sotm fall back pr

src/main/java/frc/robot/Controls.java
RobototesProgrammers: fixed

src/main/java/frc/robot/Controls.java
RobototesProgrammers: moved to drivebase bindings

jamesdooley4 (approved):
Looks good

PR #78, opened 2026-03-01
Climb

Fatblabs:


src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,396 @@
...
+    ntMotorCurrent = table.getDoubleTopic("Current (amps):").publish();
+    ntVoltage = table.getDoubleTopic("Voltage:").publish();
+  }
+
+  // Helper functions
+
+  /**
+   * sets the motor's position
+   *
+   * @param position A double representing the absolut target motor rotation, relative to the
```
markpete: [nit] spelling:  absolute

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,396 @@
...
+                            < CLIMB_POSITION_TOLERANCE;
+                      })
+                  .andThen(
+                      // This pulls the aligns the robot, then pulls up. At this stage we're
+                      // assuming the climb subsystem is already hooked on
+                      new AutoAlignCommand(), Commands.runOnce(() -> setMotorPosition(0), this))
+                  .beforeStarting(() -> climbState = ClimbState.Climbing)
+                  .finallyDo((interrupted) -> climbState = ClimbState.Idle);
+            },
+            Set.of(this)) // Requirements for the deferred command
```
markpete: **WPILib Command Requirements Mismatch (Crash Risk)**

Issue: The climbRoutine method utilizes Commands.defer and declares its requirements as Set.of(this). However, the dynamically built Command sequence includes new AutoAlignCommand(), which forcefully adds driveTrain to its requirements (lines 327-328). WPILib's DeferredCommand STRICTLY enforces that the generated command's requirements must be a defined subset of the requirements declared by defer(). Because driveTrain is missing from the initial defer requirements, WPILib will throw an IllegalArgumentException and crash the robot code at runtime when climbRoutine attempts to execute.
Suggestion: Calculate the requirements dynamically to include driveTrain safely without causing a NullPointerException (since Set.of() does not accept null elements if driveTrain happens to be explicitly turned off and initialized as null).
```
public Command climbRoutine(ClimbLevel state) {
  Set<Subsystem> deferReqs = driveTrain != null
      ? Set.of(this, driveTrain)
      : Set.of(this);

  return Commands.defer(
          () -> { ... },
          deferReqs)
      .onlyIf(() -> climbState == ClimbState.Idle);
}
```

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,396 @@
...
+              // Until it has stalled
+              if (Math.abs(climbMotor.getStatorCurrent().getValueAsDouble()) >= STALL_CURRENT
+                  && Math.abs(climbMotor.getVelocity().getValueAsDouble()) <= MAX_OMEGA) {
+                hits.incrementAndGet();
+              } else {
+                hits.set(0);
+              }
+            },
+            this)
+        .until(() -> hits.get() >= MAX_HITS)
```
markpete: **AutoZeroRoutine AtomicInteger State Leakage**

Issue: The AtomicInteger hits variable is instantiated locally exactly once during the WPILib Command composition step. If the zeroing routine starts, successfully logs some hits, gets interrupted midway (e.g., hits = 5), and is restarted later, hits will pick up right where it left off instead of starting at 0. This state leakage can cause the command to finish prematurely in the second run before truly stalling.
Suggestion: Refactor the command structure so that hits reliably resets to 0 upon initialization of the command execution sequence.
```
  return Commands.run(...)
      .until(() -> hits.get() >= MAX_HITS)
      .beforeStarting(() -> hits.set(0))
      ...
```

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,396 @@
...
+  private final TalonFX climbMotor;
+  private static final AprilTagFieldLayout APRIL_TAG_FIELD_LAYOUT = AllianceUtils.FIELD_LAYOUT;
+  private static final int BLUE_CLIMB_TAG_ID = 31;
+  private static final int RED_CLIMB_TAG_ID = 15;
+
+  // Magic Numbers
+  private static final double ROBOT_LENGTH = Units.inchesToMeters(35); // Inches
+  private static final double CLIMB_Y_OFFSET = Units.inchesToMeters(43.51); // Inches
+  private static final double GEAR_RATIO = 15;
+  private static final double CLIMB_POSITION_TOLERANCE = 0.1;
```
markpete: **Unrealistically Tight Odometry Convergence Tolerance**

Issue: The alignment tolerance TRANSLATION_TOLERANCE_METERS is strictly set to 0.01 meters (1 cm). Requiring an entire multi-system swerve drive on carpet to perfectly converge within 1 cm is extremely difficult. Unless the drivetrain module tuning (static friction feedforward) is unequivocally flawless, the robot is likely to minutely chatter around the setpoint infinitely, preventing isFinished() from ever returning true.
Suggestion: Consider loosening the tolerance to a more practical swerve accuracy distance like 0.03 or 0.05 meters, or ensure empirical tuning achieves 1cm without any limit cycling.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,413 @@
...
+  private static final double MM_ACCEL = 1;
+  private static final double MM_JERK = 5;
+
+  // Poses and trasnforms
+  private static final Transform2d frontBumperOffset =
+      new Transform2d(new Translation2d(ROBOT_LENGTH / 2, 0), Rotation2d.kZero);
+  private static final Transform2d climbOffSet =
+      new Transform2d(new Translation2d(0, CLIMB_Y_OFFSET), Rotation2d.kZero);
+
+  // Climb level positions
```
markpete: Are these numbers correct?  I don't think all of the levels are equi-distant if I recall correctly.

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,413 @@
...
+                hits.set(0);
+              }
+            },
+            this)
+        .until(() -> hits.get() >= MAX_HITS)
+        .finallyDo(
+            (interrupted) -> {
+              climbMotor.stopMotor();
+              if (!interrupted) {
+                climbMotor.setPosition(0);
```
markpete: Why not just call zeroMotor() here?

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,413 @@
...
+        return;
+      }
+      climbBumper = driveTrain.getState().Pose.transformBy(frontBumperOffset);
+      // X and Y pid calculations
+      double powerX = pidX.calculate(climbBumper.getX());
+      double powerY = pidY.calculate(climbBumper.getY());
+      // Overcome static friction if PID outputs are small
+      powerX =
+          MathUtil.clamp(
+              powerX + POWER_COEFFICIENT * Math.signum(powerX),
```
markpete: Description: Adding a static friction coefficient directly into the velocity variable calculation using Math.signum(powerX) means the resulting output power will inherently never drop exactly to zero since it continuously adds 0.05 to any non-zero positional error. This typically results in endless, fast oscillating "jitter" around the setpoint because when it crosses the target error flips, producing a hard switch back and forth. Recommendation: Set a deadband check. Only apply the static friction constant (or feedforward) if the PID error falls outside the designated translation tolerance, or utilize wpilib's ProfiledPIDController instead of manual signum corrections.

Let's trace exactly what happens with your math as the robot gets very close to the target:

Your MAX_TRANSLATIONAL_POWER is 2.0 and POWER_COEFFICIENT is 0.05.

```
powerX = MathUtil.clamp(
    powerX + POWER_COEFFICIENT * Math.signum(powerX),
    -2.0,
    2.0
);
```
Very small positive error: The PID controller outputs a tiny positive power, say 0.001. Math.signum(0.001) is 1.0. 0.001 + 0.05 * 1.0 = 0.051. clamp(0.051, -2.0, 2.0) evaluates to 0.051. The robot moves forward slightly.
Moving past the target: Because of that 0.051 power, the robot creeps forward passes the absolute 0.0 error mark and the error becomes slightly negative, say -0.001.
Very small negative error: The PID controller outputs -0.001. Math.signum(-0.001) is -1.0. -0.001 + 0.05 * -1.0 = -0.051. clamp(-0.051, -2.0, 2.0) evaluates to -0.051. The robot moves backwards slightly.
Because perfect sensors don't exist in the physical world, the error almost never reads as exactly 0.000000000, which means Math.signum will always return 1 or -1. What clamp() is doing is just making sure the robot doesn't command more than 2.0 volts when it's really far away. It doesn't tell the motor to stop when it's close.

To fix this, you instead want to check if the error is within a small "deadband" before applying the static friction.

Here is how you would write that:

```
// Keep the raw PID calculation
double powerX = pidX.calculate(climbBumper.getX());
// Only apply static friction if the error is large enough to care about
if (Math.abs(pidX.getPositionError()) > TRANSLATION_TOLERANCE_METERS) {
    powerX += POWER_COEFFICIENT * Math.signum(powerX);
} else {
    powerX = 0; // We are perfectly in the zone, stop the motors!
}
// Then clamp the max speed
powerX = MathUtil.clamp(powerX, -MAX_TRANSLATIONAL_POWER, MAX_TRANSLATIONAL_POWER);
```

markpete (changes requested):

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
Fatblabs: These are placeholders until I test on the actual robot

src/main/java/frc/robot/subsystems/ClimbSubsystem.java
```diff
@@ -0,0 +1,396 @@
...
+    climbMotor.getConfigurator().apply(configs);
+  }
+
+  // Periodic
+  @Override
+  public void periodic() {
+    // set motor position in network tables
+    ntVoltage.set(climbMotor.getMotorVoltage().getValueAsDouble());
+    ntMotorPos.set(climbMotor.getPosition().getValueAsDouble());
+    ntMotorCurrent.set(climbMotor.getStatorCurrent().getValueAsDouble());
```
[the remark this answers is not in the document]
Fatblabs: resolved

PR #80, opened 2026-03-02
adding wheelradiusCharacterization

iamawesomecat:
Add a command that automatically measures wheel radius. Ideally this gets merged quickly, no big changes that werent previously handled. Can be bound to a controller when its used.

### DIFF
--- a/src/main/java/frc/robot/util/tuning/WheelRadiusCharacterization.java
+++ b/src/main/java/frc/robot/util/tuning/WheelRadiusCharacterization.java
@@ -35,6 +35,9 @@ public class WheelRadiusCharacterization {
   public static Command wheelRadiusCharacterizationCommand(CommandSwerveDrivetrain drive) {
     SlewRateLimiter limiter = new SlewRateLimiter(WHEEL_RADIUS_RAMP_RATE);
     WheelRadiusCharacterizationState state = new WheelRadiusCharacterizationState();
+
+    ChassisSpeeds speeds = new ChassisSpeeds();
+    SwerveRequest driveRequest = new SwerveRequest.ApplyRobotSpeeds().withSpeeds(speeds);
     return Commands.parallel(
             // Drive control sequence
             Commands.sequence(
@@ -47,10 +50,7 @@ public class WheelRadiusCharacterization {
                 // Turn in place, accelerating up to full speed
                 Commands.run(
                     () -> {
-                      double speed = limiter.calculate(WHEEL_RADIUS_MAX_VELOCITY);
-                      SwerveRequest driveRequest =
-                          new SwerveRequest.ApplyRobotSpeeds()
-                              .withSpeeds(new ChassisSpeeds(0.0, 0.0, speed));
+                      speeds.omegaRadiansPerSecond = limiter.calculate(WHEEL_RADIUS_MAX_VELOCITY);
                       drive.setControl(driveRequest);
                     },
                     drive)),


src/main/java/frc/robot/util/tuning/WheelRadiusCharacterization.java
```diff
@@ -0,0 +1,112 @@
...
+                      state.positions = drive.getWheelRotations();
+                      state.lastAngle = drive.getState().Pose.getRotation();
+                      state.gyroDelta = 0.0;
+                    }),
+
+                // Update gyro delta
+                Commands.run(
+                        () -> {
+                          var rotation = drive.getState().Pose.getRotation();
+                          state.gyroDelta += Math.abs(rotation.minus(state.lastAngle).getRadians());
```
markpete: **Potential Odometry Wrap Issue**

Description: The calculation rotation.minus(state.lastAngle).getRadians() finds the un-normalized difference between two Rotation2d objects. If the underlying pose estimator (drive.getState().Pose.getRotation()) wraps its angle to mathematically stay within [-π, π], the difference will suddenly be ~2π at the wrap boundary. The Math.abs() will then inaccurately add ~2π to gyroDelta taking an invalid jump and severely ruining the characterization calculation.

Fix Suggestion: Wrap the radian difference with MathUtil.angleModulus() to guarantee you are always getting the shortest-path delta regardless of whether the angle jumps or continuously accumulates:

```
// Requires importing edu.wpi.first.math.MathUtil
state.gyroDelta += Math.abs(MathUtil.angleModulus(rotation.getRadians() - state.lastAngle.getRadians()));
```

src/main/java/frc/robot/util/tuning/WheelRadiusCharacterization.java
```diff
@@ -0,0 +1,112 @@
...
+                Commands.runOnce(
+                    () -> {
+                      limiter.reset(0.0);
+                    }),
+
+                // Turn in place, accelerating up to full speed
+                Commands.run(
+                    () -> {
+                      double speed = limiter.calculate(WHEEL_RADIUS_MAX_VELOCITY);
+                      SwerveRequest driveRequest =
```
markpete: Performance Concerns: Loop Allocation of SwerveRequest inside Commands.run

Description: In the drive control sequence, a new instance of SwerveRequest.ApplyRobotSpeeds is allocated dynamically every 20ms in the Commands.run block:

```
SwerveRequest driveRequest =
    new SwerveRequest.ApplyRobotSpeeds()
        .withSpeeds(new ChassisSpeeds(0.0, 0.0, speed));
```
Creating objects repeatedly within a high-frequency robot loop yields unnecessary garbage collection overhead on the RoboRIO. CTRE strongly recommends reusing a single pre-instantiated SwerveRequest object.

Fix Suggestion: Instantiate the request object outside the command's parallel grouping once, and simply mutate it inside the loop.

markpete (changes requested):

src/main/java/frc/robot/util/tuning/WheelRadiusCharacterization.java
```diff
@@ -0,0 +1,112 @@
...
+import frc.robot.generated.CompTunerConstants;
+import frc.robot.subsystems.drivebase.CommandSwerveDrivetrain;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+
+public class WheelRadiusCharacterization {
+  private static final double WHEEL_RADIUS_MAX_VELOCITY = 4; // Rad/Sec
+  private static final double WHEEL_RADIUS_RAMP_RATE = 0.5; // Rad/Sec^2
+  public static final double DRIVE_BASE_RADIUS =
+      Math.max(
```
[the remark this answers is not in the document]
markpete: Looks like Gemini already suggested this one. 👍

markpete (approved):

PR #84, opened 2026-03-04
autos for comp

Jetblackdragon:


src/main/java/frc/robot/subsystems/launcher/LauncherSubsystem.java
```diff
@@ -38,7 +37,7 @@ public LauncherSubsystem(Hood hood, Flywheels flywheels, TurretSubsystem turret)
   }
 
   public Command launcherAimCommand(CommandSwerveDrivetrain drive) {
-    return Commands.runEnd(
+    return Commands.run(
```
markpete: Description: launcherAimCommand was refactored from Commands.runEnd to Commands.run(...), effectively removing the explicit stowCommand() call upon termination. In AutoLogic.java, launcherCommand() is gated by .withTimeout(4.5). When this timeout triggers, the command cancels, but because launcherAimCommand lacks specific .addRequirements(hood, flywheels, turret) handling, it will leave those subsystems "hanging" with their previous motor values still active.
Recommendation:
Ensure launcherAimCommand() calls .addRequirements(...) for the associated subsystems so WPILib correctly applies default commands (like stow) on interruption.
chain an explicit stow call into the auto logic, e.g., .finallyDo(() -> s.launcherSubsystem.stowCommand().schedule())

markpete (approved):

markpete (approved):

Issue #99, opened 2026-03-13
Tune cruise velocity & acceleration of intake pivot


jamesdooley4:
> Should tune cruise velocity & acceleration based on the physical max of the intake. 

 _Originally posted by @jamesdooley4 in [#64](https://github.com/robototes/REBUILT2026/pull/64/changes#r2928706356)_

jamesdooley4:
Should be done with #110

PR #110, opened 2026-03-13
Ds missed commits

jamesdooley4:
@iamawesomecat - I think these are your changes.


iamawesomecat:
> Can't approve since I created the PR, but it looks good to me. The hood is way snappier now.

Snappier is a choice word huh

iamawesomecat:
> @iamawesomecat is #99 addressed by this PR?
> 
> And, yes, the hood is no longer constrained by software limits, and it's blown through the hardware limits too. :)

Yep!

PR #113, opened 2026-03-14
Issue 101 102

Fatblabs:
Added tunable booleans to dynamically enable and disable live tuning through advantage scope

### DIFF
--- a/src/main/java/frc/robot/util/tuning/NtTunableInteger.java
+++ b/src/main/java/frc/robot/util/tuning/NtTunableInteger.java
@@ -33,7 +33,7 @@ public final class NtTunableInteger {
 
   /** Gets the latest value from the dashboard (or the default if never changed). */
   public int get() {
-    return (int) sub.get();
+    return Math.toIntExact(sub.get());
   }
 
   public TimestampedInteger getAtomic() {


src/main/java/frc/robot/Controls.java
```diff
@@ -309,13 +309,13 @@ private void configureLauncherBindings() {
...
 
-    if (s.flywheels.TUNER_CONTROLLED) {
+    if (s.flywheels.TUNER_CONTROLLED.get()) {
       connected(launcherTuningController)
           .and(launcherTuningController.leftBumper())
           .onTrue(s.flywheels.suppliedSetVelocityCommand(() -> s.flywheels.targetVelocity.get()));
       launcherTuningController.a().whileTrue(Commands.parallel(s.indexerSubsystem.runIndexer()));
     }
-    if (s.hood.TUNER_CONTROLLED) {
+    if (s.hood.TUNER_CONTROLLED.get()) {
```
jamesdooley4: This is unlikely to have any effect, since those booleans would need to be set to 'true' between when the subsystems are created & when this code runs, which is likely on the order of microseconds or milliseconds. But it's on the right track, so I'm okay checking in as-is. The real fix is to move to auto-detecting tuner controllers being added or removed on the fly, which is out of scope of this PR.

jamesdooley4 (approved):

PR #116, opened 2026-03-14
Issue 103

Fatblabs:
added checks to see if hood is zeroed for all methods that return a command that controls position


src/main/java/frc/robot/subsystems/launcher/Hood.java
```diff
@@ -192,7 +194,8 @@ public Command autoZeroCommand() {
         .until(() -> hood.getStatorCurrent().getValueAsDouble() >= (STATOR_CURRENT_LIMIT - 1))
         .andThen(zeroHoodCommand())
         .withTimeout(3)
-        .withName("Automatic Zero hood");
+        .withName("Automatic Zero hood")
+        .onlyIf(() -> !hoodZeroed);
```
jamesdooley4: Need to be able to zero hood even if it's been previous zero'd, in order to fix it in case it loses power (which it did due to electrical issues in the turret wiring). So, I'd recommend letting this run in all cases.

jamesdooley4 (changes requested):
Need to let the hood be re-zeroed if its already been zeroed

PR #152, opened 2026-03-23
sotm-fallback

Fatblabs:
Falls back to drive base to rotate and aim to the hub if the turret starts tweaking. Can only be initiated by driver

### DIFF
--- a/AdvantageScope/Robot_2026CornerTurret/config.json
+++ b/AdvantageScope/Robot_2026CornerTurret/config.json
@@ -1,0 +1,65 @@
+{
+    "name": "2026 Corner Turret",
+    "isFTC": false,
+    "disableSimplification": false,
+    "rotations": [
+        {
+            "axis": "x",
+            "degrees": 90
+        },
+        {
+            "axis": "z",
+            "degrees": -90
+        }
+    ],
+    "position": [
+        0,
+        0,
+        0
+    ],
+    "cameras": [
+        {
+            "name": "LeftCamera",
+            "rotations": [
+                {
+                    "axis": "z",
+                    "degrees": 90
+                },
+                {
+                    "axis": "x",
+                    "degrees": 8
+                }
+            ],
+            "position": [
+                0.114,
+                0.368,
+                0.235
+            ],
+            "resolution": [
+                1280,
+                960
+            ],
+            "fov": 82
+        },
+        {
+            "name": "FrontCamera",
+            "rotations": [
+                {
+                    "axis": "y",
+                    "degrees": -15
+                }
+            ],
+            "position": [
+                0.267,
+                -0.051,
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/Controls.java
```diff
@@ -281,6 +281,40 @@ private void configureLauncherBindings() {
                     s.ledSubsystem.flashCommand(LEDSubsystem.LAUNCH_COLOR, 3, 0.2))
                 .ignoringDisable(true));
 
+    // When the back button is pressed, use the fallback
+    driverController
+        .back()
+        .whileTrue(
+            Commands.parallel(
+                s.launcherSubsystem.launcherAimCommandV2(),
+                Commands.run(() -> s.turretSubsystem.setTurretRawPosition(0)),
```
markpete: Description: The fallback aiming command runs Commands.run(() -> s.turretSubsystem.setTurretRawPosition(0)). However, the Commands.run(Runnable) factory method does not declare any subsystem requirements. Because s.turretSubsystem has a default command (rotateToTargetWithCalc() at line 449), the default command will not be interrupted. As a result, both the fallback command and the default command will run concurrently and send competing control requests to the turret motor, causing severe stuttering and failing to hold the turret at 0.
Recommendation: Include the turret subsystem as a requirement in the run method so that it properly interrupts the default command:`
Commands.run(() -> s.turretSubsystem.setTurretRawPosition(0), s.turretSubsystem)`

src/main/java/frc/robot/subsystems/launcher/LauncherSubsystem.java
```diff
@@ -86,6 +97,11 @@ public Command zeroSubsystemCommand() {
     return s.hood.zeroHoodCommand();
   }
 
+  public boolean isDriveBaseAtTarget() {
+    return Math.abs(launchParameters.targetTurret().getDegrees())
```
markpete: Need a null check on launchParameters to make sure it's not null.

Description: The method isDriveBaseAtTarget() accesses launchParameters.targetTurret() without verifying if launchParameters is null. launchParameters is initially null until launcherAimCommandV2 successfully executes. While isAtTargetFallback() implements a null check before calling isDriveBaseAtTarget(), isDriveBaseAtTarget() is a public method. Any direct invocation early in the robot lifecycle (or before targeting data is initialized) will result in a NullPointerException and cleanly crash the main robot loop.
Recommendation: Add a null check to immediately return false if launchParameters is not initialized.

```
public boolean isDriveBaseAtTarget() {
  if (launchParameters == null) return false;
  return Math.abs(launchParameters.targetTurret().getDegrees())
      <= TurretSubsystem.TURRET_DEGREE_TOLERANCE;
}
```

PR #164, opened 2026-03-28
Issue 157: Add NT / WPILog logging of feeder and spindexer key values

### LINKED ISSUE #157: Feeder and spindexer have no logging of system behavior
From the logfile during a testing session today at [FRC_20260325_032754.zip](https://bsd405.sharepoint.com/:u:/r/sites/SammamishRoboticsTeam2412/Shared%20Documents/Software%20Team/2026%20Practice%20Logs/FRC_20260325_032754.zip), it looks like there are no signals of the feeder or spindexer's current velocity or current draw - just the target RPS values. In this session the .hoot logs weren't recorded or kept, so there's no record of how those subsystems performed while they were tested. :(

This issue tracks adding logging for at least the velocity, stator current, and supply current for spindex & feeder. NT logging isn't needed can just use [DataLog](https://docs.wpilib.org/en/stable/docs/software/telemetry/datalog.html#custom-data-logging-using-datalog). But NT logging for some (say, velocity, that might be useful on the dashboard) and DataLog logging for the rest might be a good balance.

Fatblabs:


Fatblabs:
#157

src/main/java/frc/robot/subsystems/index/Feeder.java
```diff
@@ -92,4 +112,14 @@ public void simulationPeriodic() {
     motorSim.setInput(feedMotor.getSimState().getMotorVoltage());
     motorSim.update(TimedRobot.kDefaultPeriod);
   }
+
+  @Override
+  public void periodic() {
+    StatusSignal.refreshAll(statorCurrent, supplyCurrent);
```
markpete: Suggest adding feederRPS to this list instead of refreshing it seperately.

Description: In the periodic() loop, feederRPS.refresh() is called individually right after StatusSignal.refreshAll(statorCurrent, supplyCurrent) on line 118. This defeats the primary performance benefit of refreshAll(), which is to synchronize timestamps lock-step and optimize CAN bus traffic by bundling signal queries. Calling .refresh() separately on feederRPS will cause it to query its cache or the bus out of sync with the current signals. Recommendation: Add feederRPS to the refreshAll() parameters, and just call .getValueAsDouble() in the NetworkTables logging line without the redundant .refresh().

// Recommended Fix:
```
StatusSignal.refreshAll(statorCurrent, supplyCurrent, feederRPS);
// Log on NT at all times
SmartDashboard.putNumber("FeederSubsystem/VelocityRPS", feederRPS.getValueAsDouble());
```

markpete (dismissed):

jamesdooley4 (commented):
Just need to fresh the velocity values along with the current values, to make most efficient use of the CAN bus

jamesdooley4 (approved):

PR #165, opened 2026-03-29
3/26 autos

Jetblackdragon:


src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -153,6 +164,37 @@ public Trigger atTargetVelocityTrigger(double targetRPS, double toleranceRPS) {
     return new Trigger(() -> atTargetVelocity(targetRPS, toleranceRPS));
   }
 
+  public boolean stoppedShooting(double durationSeconds) {
```
markpete: The method is named stoppedShooting(), but the logic actually detects when the flywheels have been at target velocity for a sustained duration — it does not detect that shooting has stopped. In fact, being at target velocity is the state where shooting is actively occurring (the flywheel is spun up and held at set speed). The name implies the opposite of its behavior and will confuse future maintainers. The commented-out usage in AutoLogic.java (line 219) supports this: .until(() -> s.flywheels.stoppedShooting(2)) was apparently intended to end the launcher command after the flywheel has been at target for 2 seconds, which is a "done shooting" detection. But the name stoppedShooting implies the flywheel has slowed down, not that it has been held at speed.
Recommendation:
Rename to something like hasBeenAtTargetFor(double durationSeconds) or atTargetForDuration(double durationSeconds) to accurately describe its behavior. Update the commented-out call in AutoLogic.java if/when the method is activated.

src/main/java/frc/robot/Controls.java
```diff
@@ -179,6 +180,10 @@ private void configureDrivebaseBindings() {
       return;
     }
 
+    driverController
+        .y()
```
markpete: WheelRadiusCharacterization.wheelRadiusCharacterizationCommand() is bound to the driver controller's Y button via whileTrue(). This is a characterization/tuning routine that spins all wheels in a controlled pattern to measure wheel radius, and is not a valid in-match command. It is placed inside configureDrivebaseBindings() with no guard, so it will be active during both teleop and autonomous. In a competition setting, if a driver accidentally presses Y, the drivetrain will enter characterization mode — a significant safety risk and disqualification risk during a match.
Additionally, this binding conflicts with the existing launcherTuningController.y() binding at line 315, which sets flywheel velocity to 60 RPS. More critically, this characterization command replaces what was previously (per the old commented-out sysId bindings at lines 204–207) a sysId binding on Y. The new binding shadow-replaces sysId but is placed before the default drive command is set, which is fine positionally, but the risk of accidental activation remains.
Recommendation:
Either:
Move this binding behind a RobotBase.isReal() + Connected tuning-controller guard, or
Restrict it to a dedicated tuning controller (not the driver controller), or
Remove it before competition and only add it for characterization sessions
// Safer pattern — only active with a separate tuning controller connected:
```
connected(tuningController)
    .and(tuningController.y())
    .whileTrue(WheelRadiusCharacterization.wheelRadiusCharacterizationCommand(s.drivebaseSubsystem));
```

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -213,6 +216,7 @@ public static Command launcherCommand() {
             s.launcherSubsystem.launcherAimCommandV2(),
             Commands.waitUntil(() -> s.launcherSubsystem.isAtTarget())
                 .andThen(Commands.parallel(s.indexerSubsystem.runIndexer())))
+        //     .until(()-> s.flywheels.stoppedShooting(2));
```
markpete: Uncomment or remove dead code

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -60,9 +60,12 @@ public enum StartPosition {
           new AutoPath("C-Outpost-Depot", "C-Outpost-Depot"),
           new AutoPath("LeftTrench-Depot", "LeftTrench-Depot"),
           new AutoPath("LT-Neutral-Depot", "LT-Neutral-Depot"),
+          new AutoPath("LT-Neutral", "LT-Neutral"),
           new AutoPath("LT-DoubleNeutral", "LT-DoubleNeutral"),
           new AutoPath("RightTrench-Outpost", "RightTrench-Outpost"),
           new AutoPath("RT-Neutral-Outpost", "RT-Neutral-Outpost"),
+          new AutoPath("Rotate-RT-Neutral", "Rotate-RT-Neutral"),
```
markpete: This auto is in the WIP folder, but registered here as a competition auto.  Should this be removed or should we move this auto to the COMP folder?

src/main/java/frc/robot/generated/CompTunerConstants.java
```diff
@@ -87,7 +87,7 @@ public class CompTunerConstants {
 
   private static final double kDriveGearRatio = 5.357142857142857;
   private static final double kSteerGearRatio = 18.75;
-  public static final Distance kWheelRadius = Inches.of(1.84);
+  public static final Distance kWheelRadius = Inches.of(2.014);
```
markpete: This PR updates kWheelRadius from 1.84" to 2.014" (~51.16 mm) in CompTunerConstants.java. However, pathplanner/settings.json still has "driveWheelRadius": 0.051 (51 mm ≈ 2.008"), which is slightly inconsistent with the new value. While the PathPlanner simulation uses its own settings, odometry-based path following uses the wheel radius from CompTunerConstants. The gap is small (~0.006"), but:
It will cause the PathPlanner simulation to compute slightly different velocities and accelerations than the robot actually executes, making simulation less useful for validation.
If the change from 1.84" to 2.014" reflects a measured characterization result (likely, given the Wheel Radius Characterization binding added in Controls.java), the PathPlanner settings should be updated consistently.
Recommendation:
Update settings.json to match the new wheel radius:
// src/main/deploy/pathplanner/settings.json line 18
"driveWheelRadius": 0.05115,   // 2.014 inches = 0.051155 m
Also consider whether kSpeedAt12Volts (currently 5.30 m/s) should be recalculated with the new wheel radius if it was previously derived from odometry data.

markpete (dismissed):

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -152,6 +163,33 @@ public Trigger atTargetVelocityTrigger(double targetRPS, double toleranceRPS) {
...
+      // if no active timer
+      if (timeEnteredTargetZone < 0) {
+        // First time at target, record the timestamp.
+        timeEnteredTargetZone = Timer.getFPGATimestamp();
+        return false;
+      }
+    } else {
+      // Not at target, reset the timer to -1
+      timeEnteredTargetZone = -1;
+    }
```
jamesdooley4: This would be cleaner like this:
```suggestion
    if (!atTarget) {
        timeEnteredTargetZone = -1;
        return false;
    }

    if (timeEnteredTargetZone < 0) {
        timeEnteredTargetZone = Timer.getFPGATimestamp();
        return false;
    }
```

src/main/java/frc/robot/Controls.java
Jetblackdragon: resolved in another branch

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
jamesdooley4, line 215: Need to call `resetFuelCheck()` here in order to reset the debouncers used by `flywheels.isOutOfFuel()`, otherwise the second time this command is called the `isOutOfFuel()` will be immediately true.

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
jamesdooley4: Since `flywheelOneRPS` was added, this might as well us it. But it'd also need to explicitly refresh the signal first, since that's currently implicitly done inside the `getVelocity()` implementation. And then might as well update the getSupplyCurrent() to also cache the StatusSignal<> at the class level and refresh it in a single call. 
Not blocking, as the code as-is works due to the implicit refresh in `getVelocity()` causing the value in `flywheelOneRPS` to be updated (since it's the same Java object).

jamesdooley4 (dismissed):
Missing a call to reset the debouncer, but I see it was disabled in a separate branch. And other comment is a future improvement, since this code works as-is (albeit, a bit unexpectedly).

RobototesProgrammers (approved):

PR #167, opened 2026-03-30
Fix Issue 166: DataLog overrunning buffers, causing wpilog files to abort mid-match

### LINKED ISSUE #166: DataLog overrunning buffers, causing wpilog files to abort mid-match
The last event in the .wpilog file for WASAM_E14 happens right about the same time the DSEvents log from the Driver Station has this:

Timestamp | DSEvents
-- | --
278.980  | "DataLog: outgoing buffers exceeded threshold, pausing logging--consider flushing to disk more frequently (smaller period)"

This is referenced on Chief Delphi [here](https://www.chiefdelphi.com/t/advantagescope-2024-log-analysis-robot-telemetry-continued/442658/62) and [here](https://www.chiefdelphi.com/t/advantagekit-2024-log-replay-again/442968/163) and in an AdvantageKit issue [here](https://github.com/Mechanical-Advantage/AdvantageKit/issues/87)

This issue tracks:
* Updating code this to avoid the issue
  * Flush more frequently
  * Log less data to the .wpilog file. `Telemetry::telemeterize()` is a large offender, since it logs large structures at 250Hz to NetworkTables, causing large amounts of network traffic + lots of .wpilog logging (since it always logs changes to NetworkTables). That function already writes the same data to the more-efficient .hoot log file via `SignalLogger`, so the duplicate NetworkTables data can either be removed entirely (if it's not actively being used), or at least reduced to a reasonable rate like 25Hz or 50Hz
  * Resume logging if it's paused due to buffer exhaustion, so events after that point aren't permanently lost
* Verify USB keys used for Rio logging can be written at the Rio's full speed. Should empirically test the key write speeds when they are plugged into the Rio, and clearly mark the ones that are good to use for logging

Fatblabs:


src/main/java/frc/robot/Telemetry.java
```diff
@@ -137,17 +136,32 @@ public void telemeterize(SwerveDriveState state) {
...
+        m_moduleDirections[i].setAngle(state.ModuleStates[i].angle);
+        m_moduleSpeeds[i].setLength(state.ModuleStates[i].speedMetersPerSecond / (2 * MaxSpeed));
 
-      SmartDashboard.putData("Module " + i, m_moduleMechanisms[i]);
+        SmartDashboard.putData("Module " + i, m_moduleMechanisms[i]);
+      }
+      /* Telemeterize the pose to a Field2d */
+      fieldTypePub.set("Field2d");
+      fieldPub.set(m_poseArray);
+      cacheLoopcount = 0;
```
jamesdooley4: Can all of this happen in the main command loop and not the time-critical odometry call here? While this change does decrease the logging rate, it still makes every `currentFreqDivider` odometry loop slower than the rest, potentially affecting the odometry correctness.

jamesdooley4 (commented):
While this does reduce the logging rate, it still likely means every Nth odometry loop will be slower than the rest.

markpete (dismissed):

src/main/java/frc/robot/util/simulation/DrivebaseSim.java
```diff
@@ -0,0 +1,133 @@
...
+  public DrivebaseSim(double MaxSpeed) {
+    SwerveDriveState state = Telemetry.returnDriveState();
+    /* Telemeterize the module states to a Mechanism2d */
+    for (int i = 0; i < 4; ++i) {
+      m_moduleSpeeds[i].setAngle(state.ModuleStates[i].angle);
+      m_moduleDirections[i].setAngle(state.ModuleStates[i].angle);
+      m_moduleSpeeds[i].setLength(state.ModuleStates[i].speedMetersPerSecond / (2 * MaxSpeed));
+
+      SmartDashboard.putData("Module " + i, m_moduleMechanisms[i]);
+    }
```
jamesdooley4: `state` may be null when the programs starts if `telemeterize()` hasn't been called yet, so this needs to handle that case without crashing.

src/main/java/frc/robot/util/simulation/DrivebaseSim.java
```diff
@@ -0,0 +1,133 @@
...
+    driveTimestamp.set(state.Timestamp);
+    driveOdometryFrequency.set(state.OdometryPeriod == 0 ? 0 : 1.0 / state.OdometryPeriod);
+    /* Telemeterize the pose to a Field2d */
+    fieldTypePub.set("Field2d");
+
+    double[] poseArr = Telemetry.returnPoseArray();
+    if (poseArr != null) {
+      fieldPub.set(poseArr);
+    }
+  }
```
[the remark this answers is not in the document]
Fatblabs: image

jamesdooley4 (changes requested):
Should be able to make the Telemetry class accesses non-static, and seems to be missing the mechanism2d updates.

jamesdooley4 (changes requested):
Few minor comments, and suggestions on better class names & folder locations since these are neither sim related nor what I think of as "telemetry"

jamesdooley4 (approved):
Testing today showed this had the intended effect (logging resumes after stalling, and dramatically reduced datarate of the DriveState values and the Module2d version of swerve positions.

PR #168, opened 2026-03-30
Oscillate spindexer

Fatblabs:
need to test this first before this can be merged to the main branch. Also need mentor / aryaa review


src/main/java/frc/robot/subsystems/index/Spindexer.java
```diff
@@ -33,6 +34,13 @@ public class Spindexer extends SubsystemBase {
 
   private final FlywheelSim motorSim;
 
+  // Oscillate settings
+  private static final double PHASE_SHIFT = 0;
```
Jetblackdragon: What's the point of subtracting 0

Jetblackdragon (changes requested):

jamesdooley4 (changes requested):
The target speed is way too low. And I'm not sure why it'd be a good thing for the spindexer's velocity setpoint to oscillate, or what problem this is trying to solve.

RobototesProgrammers:
no longer needed

PR #172, opened 2026-04-02
updated hood protection + updated intake oscillation + updated launcher v3 data points

Jetblackdragon:


jamesdooley4 (approved):
Still has some work, but at a good spot to integrate to main and continue working tonight.

src/main/java/frc/robot/subsystems/launcher/LaunchCalculator.java
```diff
@@ -265,10 +267,21 @@ public double derivativeOfTOF(double distance) {
...
+    for (int i = 0; i <= TRENCH_LOOKAHEAD_SAMPLES; i++) {
+      double t = TRENCH_LOOKAHEAD * i / TRENCH_LOOKAHEAD_SAMPLES;
+      Pose2d sampledRobotPose =
+          robotPose.exp(
+              new Twist2d(
+                  speeds.vxMetersPerSecond * t,
+                  speeds.vyMetersPerSecond * t,
+                  speeds.omegaRadiansPerSecond * t));
+      Pose2d sampledTurretPose = sampledRobotPose.transformBy(turretTransform);
+      if (isCloseToTrench(sampledTurretPose)) {
```
markpete: Should we do isCloseToTrench(sampledRobotPose) and move this up?  Please add a comment on why you need the turret position for this at all.  It seems like if you're only using it to determine if you're close to the trench, you would only need the robot position and not the turret position.

src/main/java/frc/robot/subsystems/intake/IntakePivot.java
```diff
@@ -132,8 +133,8 @@ public void restartTimer() {
   }
 
   public void oscillatePivot() {
-    double offset = (1 + Math.cos(2 * Math.PI * (timer.get() / 3) + Math.PI)) / 2;
-    double pos = DEPLOYED_POS + offset;
+    double offset = (1 + Math.cos(2 * Math.PI * (timer.get() / 2) + Math.PI)) / 2;
+    double pos = LAUNCH_POS_OUT + offset * (LAUNCH_POS_IN - LAUNCH_POS_OUT);
```
markpete: The oscillation calculates pos = LAUNCH_POS_OUT + offset * (LAUNCH_POS_IN - LAUNCH_POS_OUT).

LAUNCH_POS_OUT = -0.3, LAUNCH_POS_IN = -0.18, EXTAKE_POS = -0.30 (same as LAUNCH_POS_OUT).
offset ranges from 0 to 1, so pos ranges from LAUNCH_POS_OUT (-0.3) to LAUNCH_POS_IN (-0.18). This is correct in isolation.
However, EXTAKE_POS is also -0.30, meaning the oscillation passes through the extake position during every cycle. This may cause unexpected sensor/game-piece interactions during launch.
More critically, the new formula is only correct when LAUNCH_POS_OUT < LAUNCH_POS_IN numerically. If tuning changes the relationship, the oscillation will invert silently.
Recommendation:

Add a static assertion or a Robot.isSimulation() check at startup to verify LAUNCH_POS_OUT < LAUNCH_POS_IN numerically (since positions are negative rotations, confirm the intended physical direction).
Rename EXTAKE_POS or change its value if it should not overlap with LAUNCH_POS_OUT.
Add a comment documenting the direction convention (negative = deployed/down).

markpete (changes requested):

src/main/java/frc/robot/subsystems/launcher/LaunchCalculator.java
```diff
@@ -265,21 +269,32 @@ public double derivativeOfTOF(double distance) {
...
-   * Checks to see if a point's x value (in an x,y field) on the robot is close to the trench or
-   * not.
-   *
-   * @param pose the point (pose2d) on the robot you would like to check
-   * @return True if close, false if not close
-   */
-  public static boolean isCloseToTrench(Pose2d pose) {
-    double nearestTagX = pose.nearest(trenchTags).getX();
-    return Math.abs(nearestTagX - pose.getX()) < TURRET_TO_TRENCH_TOLERANCE;
+  public boolean isApproachingTrench(Pose2d robotPose, ChassisSpeeds speeds) {
```
jamesdooley4: This can be static since it doesn't reference any instance variables - just the `static final` variable `turretTransform` and the method parameters

jamesdooley4 (approved):
Still working on datapoints, but might be nice to get in so other branches can built upon it.

PR #185, opened 2026-04-05
Issue #177: SmartDashboard/NT Writes Every Loop in TurretSubsystem.periodic()

### LINKED ISSUE #177: SmartDashboard/NT Writes Every Loop in TurretSubsystem.periodic()
src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
Lines: 268–273

Description
```
SmartDashboard.putNumber("Turret/Position", turretMotor.getPosition().getValueAsDouble());
SmartDashboard.putNumber("Turret/Target", targetPos);
SmartDashboard.putNumber("Turret/Velocity", turretMotor.getVelocity().getValueAsDouble());
SmartDashboard.putNumber("Turret/Current", turretMotor.getStatorCurrent().getValueAsDouble());
```
SmartDashboard.putNumber() is called four times per 20 ms loop, plus three TalonFX.get*().getValueAsDouble() CAN reads. While these don't directly consume significant power, every get*() call on a Phoenix 6 device that hasn't had its signal frequency configured will trigger a deferred CAN request and add latency/bus overhead.

Additionally, in Telemetry.java, SmartDashboard.putData("Module " + i, m_moduleMechanisms[i]) is called 4× per loop (every 20 ms) for Mechanism2d objects (line 150). Mechanism2d putData serializes the entire widget graph every call, which is heavier than a simple number publish.

Recommendation
Use pre-cached NetworkTables publishers (as already done in Flywheels.java and Hood.java) instead of SmartDashboard.putNumber() for the turret values.
Move the SmartDashboard.putData("Module " + i, ...) in Telemetry.java to robot initialization (done once) rather than inside telemeterize().
Use Phoenix 6's BaseStatusSignal.refreshAll() pattern to batch-refresh signals at the start of periodic() rather than fetching individually.

Fatblabs:
Fixes #177

### DIFF
--- a/src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
@@ -239,7 +239,7 @@ public class TurretSubsystem extends SubsystemBase {
    * @return the angular velocity of the turret in rad/s
    */
   public double getOmega() {
-    return Units.rotationsToRadians(velocitySignal.refresh().getValueAsDouble());
+    return Units.rotationsToRadians(velocitySignal.getValueAsDouble());
   }
 
   // Experimental IF WITHIN BOUNDS go to it instead of clamping
--- a/src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
@@ -22,7 +22,6 @@ import edu.wpi.first.units.measure.Angle;
 import edu.wpi.first.units.measure.AngularVelocity;
 import edu.wpi.first.units.measure.Current;
 import edu.wpi.first.units.measure.Voltage;
-import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
 import edu.wpi.first.wpilibj2.command.Command;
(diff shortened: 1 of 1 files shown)


jamesdooley4 (dismissed):
Looks correct. Just a couple suggestions to make it even better.

src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
```diff
@@ -208,7 +239,7 @@ public boolean atTarget() {
    * @return the angular velocity of the turret in rad/s
    */
   public double getOmega() {
-    return Units.rotationsToRadians(turretMotor.getVelocity().getValueAsDouble());
+    return Units.rotationsToRadians(velocitySignal.refresh().getValueAsDouble());
```
jamesdooley4: Shouldn't need the `.refresh()` since `TurretSubsystem.periodic()` is guaranteed to be called each loop before any commands are executed, but not blocking since this isn't any worse than before & cleaning up the subsystem to use the now-cached signals can be a different PR.

jamesdooley4 (dismissed):

jamesdooley4 (approved):

PR #192, opened 2026-04-06
Issue 176: Reduce auto zero voltage for intake pivot to 4 volts

### LINKED ISSUE #176: Intake Pivot Auto-Zero Uses Stator Stall Detection; Motor Runs at 8 V Against Hard Stop
src/main/java/frc/robot/subsystems/intake/IntakePivot.java
Lines: 30, 179–194

Description
```
private static final double AUTO_ZERO_VOLTAGE = 8;
// ...
voltageControl(() -> Volts.of(AUTO_ZERO_VOLTAGE))
    .withDeadline(
        Commands.waitSeconds(0.5)
            .andThen(Commands.waitUntil(
                () -> pivotMotor.getStatorCurrent().getValueAsDouble()
                    >= (STATOR_CURRENT_LIMIT - 1))));
```
8 V (~67% of 12 V) is applied to drive the pivot into its hard stop. The stator limit is 60 A, so the motor will be absorbing up to 480 W of power while stalled against the stop. Until the stator limit engages (after the 0.5 s delay), this is an uncontrolled stall. Even at rated limits this is a significant battery draw event.

Recommendation
Reduce AUTO_ZERO_VOLTAGE to 3–4 V — enough to reliably find the hard stop without excessive current.
Consider using a hardware limit switch for zeroing to avoid this stall pattern entirely.
Match the Hood subsystem pattern (which uses -3 V at line 62 of Hood.java) — already much better than 8 V.
java
private static final double AUTO_ZERO_VOLTAGE = 3; // was 8

Fatblabs:


markpete:
> The reason it was 8 was because it needs to go to the retracted position quickly over the bumper and the number I tested and worked was 8 im unsure if 4 will be enough -Aarya

@Jetblackdragon - Concern is that this may be causing our brown outs.  You're the prog lead, so it's your call whether we change this or leave it at 8.  Please see the linked issue for power consumption concern details.

Jetblackdragon:
I read over the issue. I think we should fine tune this at dcmps or after and just wait on the merge. Im not too worried about brownouts because this shouldnt even need to be ran during matches ideally

markpete (approved):
Approved by me, but you should get verbal approval from @Jetblackdragon before merging.  If he disagrees, you should close this without merging.

PR #194, opened 2026-04-06
Issue 118: Don't turn turret unless robot has moved significantly at least once

### LINKED ISSUE #118: Safety issue: Turret moves 180 degrees on robot start without any operator input
I've noticed this for a while, and finally filing an issue for it.

Repro:
Zero physical mechanisms
Turn on robot
Wait for code to start
Enable robot

Result:
Turret zooms around 180 degrees, likely because it's going from its initial position (pointing backwards) to pointing towards where it thinks the hub is based on the robot's default position (parked right in front of the hub with the intake pointing away from the hub).

Expected result:
Nothing on the robot moves when the robot is enabled, until some operator input happens. Or at least is moves a bit less on startup so there's less of a risk of breaking the cable chain or hitting a human that's nearby and not expecting the turret to whip around.

This may become less of an issue when the 'zero' position is at the new hard stop. Unless the zero position is chosen to be 320 degrees away from the current "aim at hub" default, so it actually gets worse than before.

Fatblabs:


src/main/java/frc/robot/subsystems/launcher/LaunchCalculator.java
```diff
@@ -107,9 +112,19 @@ public LaunchingParameters getParameters(
...
                 <= MIN_ROTATION_TOLERANCE;
-    // Check to see if all conditions are met
+
+    // On robot startup / entering enabled, check to see if it has moved at least once
+    if (!hasMovedAtLeastOnce) {
+      // has moved enough?
+      if (!hasNotMovedSignificantly || !hasNotRotatedSignificantly || !isNotMovingFastEnough) {
+        hasMovedAtLeastOnce = true;
+      } else {
+        return defaultParam; // Still haven't moved, so return null
```
markpete: Description: The new block introduces return defaultParam; but the newly added inline comment specifically states // Still haven't moved, so return null. Returning a LaunchingParameters object forcefully filled with 0 targeting fields (as defined on line 62) instead of returning null could unintentionally command the turret and hood to abruptly snap to a 0 degree setpoint on start-up. In WPILib loops, abrupt state jumps to 0 are generally undesirable. Recommendation: If the downstream commands handle a null value as a directive to bypass mechanism movement (which aligns with the comment's phrasing), replace return defaultParam; with return null;. However, if you must safely return a zeroed object, consider capturing and using the current measured positions of the turret mechanisms as the "default" so the robot holds its position rather than unexpectedly snapping to 0.

src/main/java/frc/robot/subsystems/launcher/LaunchCalculator.java
```diff
@@ -57,6 +58,10 @@ public static LaunchCalculator getInstance() {
   private static final List<Pose2d> trenchTags = new ArrayList<>();
   private static final int[] tags = {1, 6, 7, 12, 17, 22, 23, 28}; // Trench tags
 
+  // dont move launch param
+  private static final LaunchingParameters defaultParam =
```
markpete: By Java convention static variables should be in UPPER_SNAKE_CASE rather than camelCase.

defaultParam ==> DEFAULT_PARAM
even better DEFAULT_LAUNCH_PARAM

markpete (changes requested):

src/main/java/frc/robot/subsystems/launcher/LaunchCalculator.java
```diff
@@ -107,9 +112,19 @@ public LaunchingParameters getParameters(
                 <= MIN_VELOCITY_TOLERANCE
             && Math.abs(currentSpeeds.omegaRadiansPerSecond - lastSpeeds.omegaRadiansPerSecond)
                 <= MIN_ROTATION_TOLERANCE;
-    // Check to see if all conditions are met
+
+    // On robot startup / entering enabled, check to see if it has moved at least once
+    if (!hasMovedAtLeastOnce) {
+      // has moved enough?
+      if (!hasNotMovedSignificantly || !hasNotRotatedSignificantly || !isNotMovingFastEnough) {
```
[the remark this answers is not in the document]
Fatblabs: if it hasn't NOT moved

PR #195, opened 2026-04-06
Out of Memory issue launching simulator

markpete:
There have been reports of out-of-memory errors when booting the simulator app.  This PR makes two changes to attempt to address this issue:

1) When running the simulator, we allocate more memory to the process on launch.
2) Reduce the vision from a resolution of 960x720 to 320x240.  In the simulator vision is streaming via the Network Tables the images from 3 cameras simultaneously.  By reducing the resolution, the memory cost of this action is reduced by 90%.

### DIFF
--- a/src/main/java/frc/robot/sim/visionproducers/VisionSimConstants.java
+++ b/src/main/java/frc/robot/sim/visionproducers/VisionSimConstants.java
@@ -31,7 +31,7 @@ public class VisionSimConstants {
   public static final double kCameraFPS = 15;
   public static final double kAvgLatencyMs = 50;
   public static final double kLatencyStdDevMs = 15;
-  public static final double kMinTargetAreaPixels = 10.0;
+  public static final double kMinTargetAreaPixels = 1.1;
   public static final double kMaxSightRangeMeters = 3.0;
 
   /** Per-camera configuration for vision simulation. */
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/sim/visionproducers/VisionSimConstants.java
```diff
@@ -23,11 +23,11 @@ public class VisionSimConstants {
...
   // Camera simulation properties (shared by all cameras — all cameras are identical)
-  public static final int kCameraResWidth = 960;
-  public static final int kCameraResHeight = 720;
+  public static final int kCameraResWidth = 320;
+  public static final int kCameraResHeight = 240;
   public static final double kCameraFOVDegrees = 90.0;
-  public static final double kCalibErrorAvg = 0.35;
-  public static final double kCalibErrorStdDev = 0.10;
+  public static final double kCalibErrorAvg = 0.12;
+  public static final double kCalibErrorStdDev = 0.035;
```
jamesdooley4, line 30: Hrm. I'd really like way to optionally have these better match the LimeLight cameras (1280x800) so it's easier to run a higher-fidelity vision simulation for people looking into improvements there. Maybe have another boolean property like `automatedTest` in build.gradle, and set a build-time variable based on that, and then switch the constants here to either the low-fidelity values. I could do that as a follow-up PR, since I agree the default should be the low-fidelity, low-memory version.

jamesdooley4 (approved):

PR #201, opened 2026-04-08
Issue 199 and 200

### LINKED ISSUE #199: Stale vision poses from throttling during auto or teleop cause robot to drive wildly in auto
Saw this during programming test time this afternoon. Root cause is the alternating MT1 / MT2 poses. Here's the summary from analyzing it with Copilot based on some of things I saw:

# Root Cause: Robot Diving Toward Red Side During Auto

**Log file:** `FRC_20260407_231323.wpilog`
**Incidents:** t≈381–385s (auto 2) and t≈530–535s (auto 3)
**Symptom:** Robot lurches toward the red side of the field during autonomous even though `DriveState/Pose` and `ModuleStates` show the wheels are commanding motion in a different direction.

---

## Summary

Stale Limelight botpose data published at 60 Hz flooded the WPILib pose estimator with hundreds of phantom vision corrections, pinning the fused pose to an outdated position while the robot was physically driving away. PathPlanner saw the robot as stationary and kept commanding velocity to catch up to its trajectory, driving the robot off-course.

---

### 1. Limelight publishes stale data at 60 Hz

LL-a's AprilTag pipeline only produced a **new frame every ~2.5 seconds**, but the Limelight hardware continued publishing the **same botpose array** to NetworkTables at its full 60 Hz rate. During the 6-second incident 1 window (379–385s):

| Metric | Value |
|--------|-------|
| Genuine new frames | 4 |
| Stale repeats | 353 |
| Ratio | 88 stale copies per real frame |

The four actual frames were:

| Time | x | y | Tags | Avg dist |
|------|---|---|------|----------|
| 379.001s | 7.128 | 7.163 | 1 | 4.89m (bad, off-field) |
| 379.530s | 3.313 | 0.448 | 1 | 1.13m (good) |
| 382.049s | 3.293 | 0.481 | 1 | 1.15m (good) |
| 384.564s | 0.000 | 0.000 | 0 | — (no tags) |

### 2. Each stale copy gets a unique timestamp

In `LLCamera.getBetterBotPoseEstimate()` (line 140):

```java
TimestampedDoubleArray tsValue = poseEntry.getAtomic();
// ...
double adjustedTimestamp = (timestamp / 1000000.0) - (latency / 1000.0);
```

The `timestamp` comes from `tsValue.timestamp` — the **NT server timestamp**, which increments every time NetworkTables publishes a value, even if the array content is identical. So each of the 353 stale copies received a **distinct `adjustedTimestamp`**, and `addVisionMeasurement()` treated every one as an independent observation.

### 3. Filters failed to catch the stale data

| Filter | Why it passed |
|--------|---------------|
| `lastFieldPose.equals(visionPose2d)` | MT1 gave (3.313, 0.448), MT2 gave (3.319, 0.428). Each clears the duplicate check for the other because `processLimelight()` is called once for MT1 and once for MT2 per camera per cycle. |
| Velocity plausibility | Stale poses have near-zero implied speed between readings — always passes. |
| Ambiguity gate (< 0.4) | The original reading was valid (1 tag at 1.13m), so all stale copies inherit the same valid ambiguity. |
| Range gate (MT1 ≤ 2m) | 1.13m < 2.0m — passes. |
| 3D sanity (height, rotation) | The stale pose had valid Z/roll/pitch — passes. |
| On-field check | (3.313, 0.448) is on-field — passes. |
| Inter-tag spread | Single-tag → spread = 0 — passes. |

**Result:** 27 accepted measurements during incident 1 from LL-a alone, all pulling the fused pose toward (3.31, 0.45).

### 4. Pose estimator is overwhelmed

The Kalman filter received ~60 "independent" vision corrections per second, all pointing to the same (stale) position. This massively outweighed the odometry updates, effectively **pinning the fused pose** near (3.3, 0.45) even as the wheels drove the robot at 3–4 m/s in the +Y direction.

### 5. PathPlanner compensates, driving the robot off-course

| Time | Fused pose | Actual (from wheels) | What happens |
|------|-----------|---------------------|--------------|
| 379–381s | (3.3, 0.5), stopped | Stationary | Stale vision flooding, fused pose pinned |
| 381.5–384s | (3.3–3.5, 0.5–0.6) | Moving +Y at 3.7 m/s | Fused pose barely moves → PP thinks robot is behind trajectory → commands more velocity toward target |
| 384.5s | Jumps to (4.7, 1.3) | ~(4.7, 1.3) | Stale data stops, pose snaps to reality |
| 385s | (4.75, 1.29), stopped | Stopped | Path ended, but robot is ~1.5m from intended position, displaced toward red side |

The PP target was (8.0, 2.9) — toward the red wall. Because the fused pose was held near (3.3, 0.5) while the robot was physically past (3.5, 0.8), the path controller over-commanded, pushing the robot further toward red than intended.

### Incident 2 (t≈530–535s)

Same pattern. LL-b accepted 7 stale measurements during this window, all reporting (6.348, 0.811) with 2 tags at yaw ≈108°. The fused heading was pulled ~20° off, causing the autonomous path to arc toward the red side.

---

### Primary: Stale-data detection in `LLCamera`

Compare the raw botpose array against the previous one. If unchanged, skip it:

```java
// In LLCamera, add per-entry tracking:
private double[] lastRawBotpose = null;
private double[] lastRawBotposeOrb = null;

// In getBetterBotPoseEstimate(), after reading poseArray:
double[] lastRaw = isMegaTag2 ? lastRawBotposeOrb : lastRawBotpose;
if (lastRaw != null && java.util.Arrays.equals(poseArray, lastRaw)) {
    return null;  // stale data — skip
}
if (isMegaTag2) {
    lastRawBotposeOrb = poseArray.clone();
} else {
    lastRawBotpose = poseArray.clone();
}
```

This eliminates the 88:1 stale-to-real ratio and ensures the Kalman filter only processes genuine new observations.

### Alternative: Compare key fields only

If exact array comparison is a concern (floating-point noise across LL firmware versions), compare just the pose fields:

```java
private boolean isStalePose(double[] current, double[] last) {
    if (last == null) return false;
    // Compare x, y, z, roll, pitch, yaw, tagCount
    for (int i : new int[]{0, 1, 2, 3, 4, 5, 7}) {
        if (Double.doubleToRawLongBits(current[i]) != Double.doubleToRawLongBits(last[i])) {
            return false;
        }
    }
    return true;
}
```

### LINKED ISSUE #200: Vision bad heading can influence robot pose heading, leading to feedback loop & wildly incorrect MT2 poses
Also saw this at programming session today:

### Use raw Pigeon2 yaw for heading supply

To fully decouple the heading feedback loop:

```java
// In Robot.supplyYawToAllLimelights():
double heading = subsystems.drivebaseSubsystem.getPigeon2().getYaw().getValueAsDouble();
// instead of:
// double heading = subsystems.drivebaseSubsystem.getState().Pose.getRotation().getDegrees();
```

This guarantees vision corrections never influence the heading sent to the Limelights.

koolpoolo:
applies fixes here


src/main/java/frc/robot/Robot.java
```diff
@@ -336,7 +336,7 @@ private void setupLimelightForAprilTags(String limelightName, boolean isEntering
 
   private void supplyYawToAllLimelights() {
     if (subsystems.visionSubsystem != null) {
-      double heading = subsystems.drivebaseSubsystem.getState().Pose.getRotation().getDegrees();
+      double heading = subsystems.drivebaseSubsystem.getPigeon2().getYaw().getValueAsDouble();
```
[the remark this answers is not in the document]
koolpoolo: wrong it provides the value in degrees

src/main/java/frc/robot/Robot.java
```diff
@@ -335,8 +335,8 @@ private void setupLimelightForAprilTags(String limelightName, boolean isEntering
   }
 
   private void supplyYawToAllLimelights() {
-    if (subsystems.visionSubsystem != null) {
-      double heading = subsystems.drivebaseSubsystem.getState().Pose.getRotation().getDegrees();
+    if (subsystems.visionSubsystem != null && subsystems.drivebaseSubsystem != null) {
+      double heading = subsystems.drivebaseSubsystem.getPigeon2().getYaw().getValueAsDouble();
```
jamesdooley4: Might need to add a call somewhere to call pigeon2.setYaw() to set the heading to a known orientation, as I'm not sure if SwerveDrivetrain.resetPose() does that or not. Else this only works if the robot is turned on when pointing at the 'zero' direction.

jamesdooley4 (commented):
Not sure what resets the gyro

src/main/java/frc/robot/Robot.java
jamesdooley4: In the simulator, the 'reset gyro' button doesn't change the Yaw in the Pigeon.

koolpoolo: just made it be subsystems.drivebaseSubsystem.getState().Pose.getRotation().getDegrees(); again

jamesdooley4 (dismissed):

Jetblackdragon:
got merged with clock and hub status so closing pr

PR #205, opened 2026-04-09
stop calling .steControl(Follower) on every single motor request

Fatblabs:


Fatblabs:
Resolved in dcmp branch

PR #220, opened 2026-04-10
Set hood position to 0 on disabledExit()

Fatblabs:


jamesdooley4 (approved):
This will change behavior when doing testing on the playground, as the hood will drop down when the robot re-enables (vs. being left at the current position). But that will only happen if a command doesn't immediately set it to something else when the next mode starts.
But I think that's a good tradeoff to avoid breaking the turret on the practice field when running two autos in a row.

PR #221, opened 2026-04-10
Flywheel idle speed

### LINKED ISSUE #218: Flywheels use stopMotor() in command end() rather than a coast/idle velocity
This may be leading to the initial shots of a volley being short as the motors are spinning up.

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
Lines: 162–165, 176–179

```
() -> {
    FlywheelOne.stopMotor();
    FlywheelTwo.stopMotor();
}
```
Root cause: stopMotor() sets duty cycle to neutral (0 V), allowing the flywheels to coast down. When the next shot command triggers, the flywheels must spin back up from (potentially) a low RPM, drawing peak current again. A better strategy is to idle the flywheels at a low RPM between shots to reduce spin-up current spikes.

Suggested fix:

```
// In Flywheels.java
public static final double IDLE_RPS = 10.0; // low idle to reduce spin-up current
// In setVelocityCommand end():
() -> setVelocityRPS(IDLE_RPS)  // idle instead of full stop
```

Fatblabs:
Repeated acceleration from an idle RPS should in theory be more energy efficient than ramping up from 0 (or near 0) every time we want to shoot

### DIFF
--- a/src/main/java/frc/robot/subsystems/launcher/Flywheels.java
+++ b/src/main/java/frc/robot/subsystems/launcher/Flywheels.java
@@ -58,6 +58,9 @@ public class Flywheels extends SubsystemBase {
   private StatusSignal<AngularVelocity> flywheelOneRPS;
   private StatusSignal<Current> flywheelOneSupplyCurrent;
 
+  // idle flywheel
+  public static final double IDLE_FLYWHEEL_RPS = 10;
+
   // Constructor
   public Flywheels() {
     flywheelOne = new TalonFX(Hardware.FLYWHEEL_ONE_ID);
--- a/src/main/java/frc/robot/subsystems/launcher/Flywheels.java
+++ b/src/main/java/frc/robot/subsystems/launcher/Flywheels.java
@@ -153,6 +153,10 @@ public class Flywheels extends SubsystemBase {
     flywheelOne.setControl(request);
   }
 
+  public void runIdleRPS() {
+    setVelocityRPS(IDLE_FLYWHEEL_RPS);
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -150,6 +153,10 @@ public void setVelocityRPS(double rps) {
     flywheelOne.setControl(request);
   }
 
+  public void runIdleRPS() {
```
markpete, line 156: Recommend private instead of public

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -58,6 +58,9 @@ public class Flywheels extends SubsystemBase {
   private StatusSignal<AngularVelocity> flywheelOneRPS;
   private StatusSignal<Current> flywheelOneSupplyCurrent;
 
+  // idle flywheel
+  public static final double IDLE_FLYWHEEL_RPS = 10;
```
markpete, line 62: Recommend private instead of public

markpete (changes requested):

Jetblackdragon:
I wouldn't want to use this. The reason we use the stop motor is that we want the motor to stop spinning without being commanded to go down to 0. The reason for this is that it would take a lot of power to slow the motors down to zero versus letting them free spin down

RobototesProgrammers:
comment says why we dont do this

PR #225, opened 2026-04-13
Dcmp

### LINKED ISSUE #181: Passing tries to shoot at a point 6 feet in the air
(The hub is the 6 foot in the air target). This is now an issue with the longer passing shots, since they will hit the driver station glass. This is also an issue with short shots, since those shoot straight up (and dont make it in the zone). This also causes excessive power draw when passing close to our own bump.

Fix: When passing, the max hood angle should always be used, and the flywheel RPM should be varied with a simple linear (maybe?) relationship to distance.

### LINKED ISSUE #206: Flywheel two-level supply current protection is disabled + missing voltage ramp
@Jetblackdragon - **Consider if we need to adjust these down or if your happy with these existing settings.**

## Summary

Two configuration bugs in `Flywheels.java` allow the flywheel motors to spike and hold
high supply current with no hardware-level throttling. This is a contributing cause of
the observed brownouts and the likely root cause of the failed flywheel motor.

**Context:** The flywheel is a two-motor (Kraken X60) primary scoring mechanism that fires
a sustained volley of up to 40–50 balls at **6–10 balls/second** (one ball every 100–167 ms).
At this rate the inter-ball window is shorter than a single flywheel recovery burst, meaning
the motors are under essentially **continuous high-load recovery** for the full 4–8 second volley.

---

## Affected File

`src/main/java/frc/robot/subsystems/launcher/Flywheels.java` — `configureMotors()`, lines 93–97

---

### Current code
```java
config.CurrentLimits.SupplyCurrentLimit      = 80;
config.CurrentLimits.SupplyCurrentLimitEnable = true;
config.CurrentLimits.SupplyCurrentLowerLimit  = 0;   // ← BUG: disables two-level protection
// SupplyCurrentLowerTime not set — defaults to 0 s
config.CurrentLimits.StatorCurrentLimit       = 100;
config.CurrentLimits.StatorCurrentLimitEnable = true;
```

### Root cause

Phoenix 6's two-level supply limiting is described in the
[CTRE docs](https://v6.docs.ctr-electronics.com/en/stable/docs/hardware-reference/talonfx/improving-performance-with-current-limits.html):

> *"if the supply current limiter has been actively limiting for the `SupplyCurrentLowerTime`,
> the limit will be reduced to the `SupplyCurrentLowerLimit` until supply current is below
> that limit, preventing breakers from tripping."*

Setting `SupplyCurrentLowerLimit = 0` means the throttle-back threshold is **never crossed**
(supply is always ≥ 0 A), so the protection **never activates**. Both motors can sustain 80 A
supply indefinitely — directly contributing to bus sag and motor thermal damage during a
multi-second volley.

`SupplyCurrentLowerTime` is also unset and defaults to `0 s`. Even with a corrected lower limit,
a `LowerTime` of 0 s would throttle supply immediately, starving recovery torque between shots.

### Why stator limits should NOT be reduced

`StatorCurrentLimit = 100 A` is intentionally preserved:

> *"Stator current limits also restrict acceleration. It is important not to set stator current
> limits too low, otherwise performance may be negatively impacted."* — CTRE docs

At 6–10 balls/second, stator current directly controls velocity recovery between each ball.
Reducing stator reduces recovery torque, which at this firing rate means velocity dips
**accumulate across the volley** rather than clearing between shots. The flywheel's rotational
inertia does most of the per-shot work; the motor's job is to continuously replenish that
kinetic energy. Starving stator current undermines the entire mechanism.

### Choosing `SupplyCurrentLowerLimit` and `SupplyCurrentLowerTime` at 6–10 balls/second

At 6–10 balls/second the motor has no meaningful idle time between shots. Supply current stays
elevated for essentially the entire volley (~4–8 seconds for 40–50 balls). This changes how
the two-level window behaves:

| `SupplyCurrentLowerTime` | Effect at 10 balls/sec |
|---|---|
| 0.5 s | Throttle engages after ~ball 5. Most of the volley runs at the lower limit. |
| 1.5 s | Throttle engages after ~ball 9–15. First half of volley at full peak; second half at sustained limit. |
| 4.0+ s | Throttle never engages during a 40-ball volley. No sustained protection. |

**`SupplyCurrentLowerTime = 1.5 s`** is the right balance: it allows the full 80 A peak
through the critical first 1–1.5 seconds of the volley (roughly the first 9–15 balls where
the flywheel is also coming up to stable operating temperature), then transitions to the
sustained limit for the remainder.

**`SupplyCurrentLowerLimit = 70 A`** is chosen — not 60 A — because at 6–10 balls/second
the sustained limit becomes the effective operating current for most of the volley. At
70 A supply with duty cycle ≈ 0.9 (motor near full speed), available stator ≈ 70 / 0.9
≈ **78 A** — a strong recovery budget that maintains shot cadence. Dropping to 60 A would
yield ~67 A stator sustained, visibly degrading the shot rate in the second half of the volley.

---

## Bug 2: No closed-loop voltage ramp on initial spin-up

Without a `VoltageClosedLoopRampPeriod`, the PID commands full voltage instantly when the
driver presses the launch trigger. This spike coincides with the drivebase, feeder, spindexer,
and turret all activating simultaneously — the combination causes the bus voltage sag that
produces brownouts on match start.

> **Note:** This ramp only materially affects cold start (0 RPS → target). Mid-volley recovery
> bursts are a small additional voltage push from an already ~90% output state; a 0.2 s ramp
> on a ~10% swing takes ~20 ms, which does not affect shot cadence.

---

## Bug 3: Motor temperature not monitored (critical for 4–8 s sustained volley)

At 6–10 balls/second, both motors run under continuous high-load recovery for 4–8 seconds.
The TalonFX has built-in thermal protection that will silently derate output when winding
temperature becomes excessive — causing unexplained velocity drops late in the volley and,
over repeated matches, motor failure.

Currently only velocity and supply current are logged to NT. Temperature is never queried.

---

### `Flywheels.java` — `configureMotors()`

```java
private void configureMotors() {
    TalonFXConfiguration config = new TalonFXConfiguration();

    // Supply current:
    //   Peak of 80 A per motor is available for the first 1.5 s of the volley
    //   (~9–15 balls at competition firing rate). After 1.5 s of continuous draw,
    //   throttles to 70 A sustained for the remainder of the volley.
    //
    //   At 70 A supply / ~0.9 duty cycle at speed → ~78 A stator available sustained.
    //   This preserves strong velocity recovery throughout the full voley.
    config.CurrentLimits.SupplyCurrentLimit      = 80;   // peak (unchanged)
    config.CurrentLimits.SupplyCurrentLimitEnable = true;
    config.CurrentLimits.SupplyCurrentLowerLimit  = 70;   // FIX: was 0 (disabled)
    config.CurrentLimits.SupplyCurrentLowerTime   = 1.5;  // FIX: was unset (0 s)

    // Stator: keep at 100 A. At 6–10 balls/sec stator current directly controls
    // velocity recovery cadence. Reducing this degrades the shot rate.
    config.CurrentLimits.StatorCurrentLimit       = 100;  // unchanged
    config.CurrentLimits.StatorCurrentLimitEnable = true;

    // Ramp: softens the 0 → full-speed voltage spike on initial trigger press.
    // Does not affect mid-volley recovery (motor already at ~90% output).
    config.ClosedLoopRamps.VoltageClosedLoopRampPeriod = 0.20; // FIX: was unset

    // Neutral mode and invert (unchanged)
    config.MotorOutput.NeutralMode = NeutralModeValue.Coast;
    config.MotorOutput.Inverted    = InvertedValue.Clockwise_Positive;

    // PID Slot 0 — spin-up (unchanged)
    config.Slot0.kP = 0.1;

### LINKED ISSUE #212: Turret accelerates at 1200 RPS/s with no torque ramp
File: src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
Lines: 71–74

```
private static final double CRUISE_VELOCITY = 200;   // RPS
private static final double ACCELERATION    = 1200;  // RPS/s
private static final double JERK            = 6000;  // RPS/s²
```
Root cause: The turret runs at 20 A supply / 40 A stator — reasonable on its own. But during a full launch sequence the turret, hood, flywheels, feeder, and spindexer are all active at once. Any aggressive turret slew adds to the aggregate bus load. 1200 RPS/s is a very fast acceleration for a 72:1 gear ratio; the resulting torque demand at the motor is high.

Suggested fix: Reduce the turret MotionMagic acceleration to 600–800 RPS/s during normal auto-aim. The turret's large gear ratio means you get plenty of output speed at lower motor velocities:

```
private static final double ACCELERATION = 700;  // RPS/s — smoother, less bus sag
private static final double JERK         = 3000; // RPS/s²
```

### LINKED ISSUE #214: No Voltage Compensation configured on non-swerve mechanisms
Files: Flywheels.java, Feeder.java, Spindexer.java, IntakeRollers.java

Root cause: None of the non-swerve mechanisms enable VoltageCompensation or configure a VoltageCompensationConfigs entry. During brownout conditions, the output voltage delivered to the motors varies dramatically. Without voltage compensation, the velocity PID will fight harder to reach setpoint during sags, running higher than desired current exactly when the battery can least support it.

Suggested fix (example for Feeder):

```
// Add to feederConfig()
talonFXConfiguration.Voltage.PeakForwardVoltage  = 11.5; // leave headroom below 12 V
talonFXConfiguration.Voltage.PeakReverseVoltage  = -11.5;
```
This caps the maximum voltage request so the motor won't try to overdrive when battery voltage is good.

### LINKED ISSUE #215: Feeder runs at 95 RPS with no PID velocity ramp
File: src/main/java/frc/robot/subsystems/index/Feeder.java
Lines: 31, 92–93

```
private final double D_TARGET_RPS = 95;  // very high for a feeder
...
talonFXConfiguration.Slot0.kV = 11.28 / 92;
talonFXConfiguration.Slot0.kP = 0.5;
```
Root cause: 95 RPS on a single Kraken is only slightly below the motor's free-spin speed (≈ 100 RPS at 12 V). The kP = 0.5 with no ramp means a step from 0→95 RPS produces a large initial error output that pegs the motor voltage at maximum. No ClosedLoopRamps setting is configured.

Suggested fix:

```
// Reduce target RPS or add a closed-loop ramp
talonFXConfiguration.ClosedLoopRamps.VoltageClosedLoopRampPeriod = 0.15; // 150 ms ramp
private final double D_TARGET_RPS = 80; // review whether 95 RPS is actually needed
```

### LINKED ISSUE #216: Spindexer runs at 70 RPS with 1000 RPS/s acceleration
File: src/main/java/frc/robot/subsystems/index/Spindexer.java
Lines: 31–32

```
private final double D_TARGET_RPS  = 70;
private final double D_TARGET_ACCEL = 1000; // Rotations /s /s
```
Root cause: 1000 RPS/s² acceleration means the spindexer reaches full speed in 70 ms — essentially a step input on the velocity controller. Similar to the feeder issue; the PID will request maximum voltage for 70 ms on every startup, adding a current spike to the already loaded bus.

Suggested fix:
`
private final double D_TARGET_ACCEL = 300; // RPS/s — still fast but softer ramp`

### PR DESCRIPTION
**Pull request by Jetblackdragon**
missing some ds commits will be added monday

### CODE REVIEW

File: src/main/java/frc/robot/Robot.java [hunk 1]
```diff
@@ -230,10 +242,35 @@ public void disabledExit() {
...
+      if (subsystems.visionSubsystem.limelightbOnline) {
+        setupLimelightForAprilTags(Hardware.LIMELIGHT_B, true);
+        LimelightHelpers.setRewindEnabled(Hardware.LIMELIGHT_B, true);
+      }
+      if (subsystems.visionSubsystem.limelightcOnline) {
+        setupLimelightForAprilTags(Hardware.LIMELIGHT_C, true);
+        LimelightHelpers.setRewindEnabled(Hardware.LIMELIGHT_C, true);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
koolpoolo: well generally disabledInit runs before all the subsystems are initialized so we have to run it in disabled periodic

src/main/java/frc/robot/subsystems/index/IndexerSubsystem.java
```diff
@@ -24,4 +25,18 @@ public Command runIndexer() {
           spindexerSubsystem.stopMotor();
         });
   }
+
+  public Command runIndexer(DoubleSupplier flywheelRPS) {
+    return Commands.runEnd(
+        () -> {
+          double fRPS = flywheelRPS.getAsDouble() * 0.83 + 15;
+          double sRPS = Math.min(fRPS * 1.5, 70);
```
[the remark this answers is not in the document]
markpete: +1

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -93,33 +85,28 @@ private void configureMotors() {
...
-    config.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
 
     // create PID gains
-    config.Slot0.kP = 0.1;
+    config.Slot0.kP = 0.5;
     config.Slot0.kI = 0.0;
     config.Slot0.kD = 0.0;
     config.Slot0.kA = 0.0;
-    config.Slot0.kV = 6.0 / 45.0;
+    config.Slot0.kV = ((6.0 / 46.4) + (10.0 / 79.0)) / 2.0;
```
markpete: Please add a comment that tells what this math is.  I assume updated gear ratios, but a comment here will help maintainers.

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -144,60 +131,38 @@ private void applyConfig(TalonFX motor, TalonFXConfiguration config) {
...
-      request = request.withSlot(1);
-    } else {
-      request = request.withSlot(0);
-    }
-  }
-
   public Command setVelocityCommand(double rps) {
     return runEnd(
             () -> {
               request.Velocity = rps;
```
markpete: Since this code is the same as setVelocityRPS(), just replace this with a call to that method.

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -144,60 +131,38 @@ private void applyConfig(TalonFX motor, TalonFXConfiguration config) {
...
-            })
-        .withName("Set Flywheel Supplied Velocity");
-  }
-
   public void setVelocityRPS(double rps) {
     request.Velocity = rps;
-    FlywheelTwo.setControl(request);
-    FlywheelOne.setControl(follow);
+    flywheelOne.setControl(request);
+    flywheelTwo.setControl(request);
```
markpete: Why did we remove follow?

src/main/java/frc/robot/Controls.java
```diff
@@ -215,13 +216,15 @@ private void configureDrivebaseBindings() {
...
-                .runOnce(() -> s.drivebaseSubsystem.seedFieldCentric())
-                .alongWith(rumble(driverController, 0.5, Seconds.of(0.3)))
-                .withName("Reset gyro"));
+    // driverController
+    //     .back()
+    //     .onTrue(
+    //         s.drivebaseSubsystem
+    //             .runOnce(() -> s.drivebaseSubsystem.seedFieldCentric())
+    //             .alongWith(rumble(driverController, 0.5, Seconds.of(0.3)))
+    //             .withName("Reset gyro"));
```
markpete, line 226: Delete dead code

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
```diff
@@ -93,33 +85,28 @@ private void configureMotors() {
...
-
-    applyConfig(FlywheelOne, config);
-    applyConfig(FlywheelTwo, config);
+    config.MotorOutput.Inverted = InvertedValue.CounterClockwise_Positive;
+    applyConfig(flywheelOne, config);
+
+    config.MotorOutput.Inverted = InvertedValue.Clockwise_Positive;
+    applyConfig(flywheelTwo, config);
+
+    // flywheelTwo.setControl(new Follower(Hardware.FLYWHEEL_ONE_ID, MotorAlignmentValue.Opposed));
```
markpete: Remove dead code

src/main/java/frc/robot/Robot.java
```diff
@@ -230,10 +242,35 @@ public void disabledExit() {
...
+      }
+    }
   }
 
   @Override
-  public void disabledPeriodic() {}
+  public void disabledPeriodic() {
+    if (subsystems.visionSubsystem != null) {
+      if (subsystems.visionSubsystem.limelightaOnline) {
+        setupLimelightForAprilTags(Hardware.LIMELIGHT_A, true);
```
markpete, line 261: Description: The PR introduces logic inside disabledPeriodic that repeatedly calls setupLimelightForAprilTags and LimelightHelpers.setRewindEnabled. Setting Limelight properties such as SetIMUAssistAlpha, SetIMUMode, and setPipelineIndex writes to NetworkTables continuously (every 20ms the Disabled loop runs). Continuous bulk writes to NetworkTables cause heavy CPU/Network load and can lead to NT flooding or queue saturation, especially on bandwidth-constrained FRC field networks.

Recommendation: Move these configuration changes (the ones setting properties rather than just reading from them) out of disabledPeriodic(). Place them in disabledInit() so the properties are only pushed to NetworkTables once when the robot transitions to the Disabled state.

src/main/java/frc/robot/Robot.java
```diff
@@ -230,10 +242,35 @@ public void disabledExit() {
...
+      } else {
+        DriverStation.reportWarning(
+            "Attempted to set hood position although it hasn't been zeroed", false);
+      }
+    }
   }
 
   @Override
-  public void disabledPeriodic() {}
+  public void disabledPeriodic() {
```
markpete, line 258: disabledInit() calls SetThrottle() on limelights immediately but disabledPeriodic() then re-runs full setup
Lines: 210–218 (disabledInit) vs. 258–272 (disabledPeriodic)

Description:
disabledInit() sets the throttle on each camera, but disabledPeriodic() then also calls setupLimelightForAprilTags() which sets IMU mode and pipeline. The two operations partially overlap and may conflict depending on execution timing, since disabledPeriodic starts running 20 ms after disabledInit. This is an ordering and intent-clarity issue.

Recommendation:
Consolidate all disabled-mode Limelight configuration into a single well-documented method called once from disabledInit().

markpete (changes requested):

src/main/java/frc/robot/subsystems/index/IndexerSubsystem.java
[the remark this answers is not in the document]
Jetblackdragon: they are magic numbers Tristan came up with

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
Jetblackdragon: kv is different at different speeds so we average it

src/main/java/frc/robot/Controls.java
Jetblackdragon: we want it as a reference for next year because its no longer needed this year but will be needed again enxt year so I wanted to keep it

src/main/java/frc/robot/Robot.java
koolpoolo: well generally disabledInit runs before all the subsystems are initialized so we have to run it in disabled periodic so no

src/main/java/frc/robot/Robot.java
koolpoolo: well generally disabledInit runs before all the subsystems are initialized so we have to run it in disabled periodic

koolpoolo: and disabled periodic covers time in which subsystems boot up so we can't say one and done

src/main/java/frc/robot/subsystems/index/IndexerSubsystem.java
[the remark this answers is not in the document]
jamesdooley4: Not really magic - they're the ratios to get fuel to transition more smoothly between spindexer, feeder & flywheel by having it not suddenly change velocity at the transition points. But I'm not sure what you'd call them, as they're essentially the `a` and `b` in the `output = a + bx` equation. Just having a comment before each equation explaining what it does would likely be better.

src/main/java/frc/robot/subsystems/launcher/LaunchCalculator.java
```diff
@@ -129,7 +131,12 @@ public LaunchingParameters getParameters(
         && isNotMovingFastEnough
         && isTurretOmegaStable
         && cachedParams != null) {
-      return cachedParams;
+      return new LaunchingParameters(
+          cachedParams.targetHood,
+          cachedParams.targetTurret,
+          cachedParams.targetFlywheels,
+          0,
+          cachedParams.turretPose);
```
jamesdooley4: It'd be less stress on the Java garbage collector to just set `cachedParams.targetTurretFeedforward = 0` and return `cachedParams` instead of creating a new object (which will soon be thrown away and garbage collected).

src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
```diff
@@ -56,22 +55,17 @@ public class TurretSubsystem extends SubsystemBase {
...
-  private static final double kP = RobotType.isAlpha() ? 2.97 : 500;
+  private static final double kP = RobotType.isAlpha() ? 2.97 : 150;
   private static final double kI = 0;
-  private static final double kD = 0;
+  private static final double kD = RobotType.isAlpha() ? 0 : 15;
   private static final double kG = 0;
-  private static final double kS = RobotType.isAlpha() ? 0.41 : 0.346;
-  private static final double kV = 0.884766 / 1.125;
+  private static final double kS = RobotType.isAlpha() ? 0.41 : 0.36;
+  private static final double kV = 12 / 1.29; // volts per requested rps
```
jamesdooley4: Likely needs an `.isAlpha()` check so the alpha bot can have a different constant since its previous kV was a lot lower (less than one) vs the new value (around 10).

src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
```diff
@@ -169,13 +163,14 @@ public void setTurretRawPosition(double pos, double FFVelocity) {
     // KV must be converted to volts. Right now it's only in
     // dutycycle per requested rotation per second, so multiply
     // 12 to get true voltage
```
jamesdooley4: Need to update comment since kV is now in nominal volts and the extra multiply isn't needed

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -215,16 +217,14 @@ public static Command launcherCommand() {
     return Commands.parallel(
             Commands.runOnce(
                 () -> {
-                  s.flywheels.switchSlot(true);
                   s.flywheels.resetFuelCheck();
                 }),
             s.launcherSubsystem.launcherAimCommand(),
             Commands.waitUntil(() -> s.launcherSubsystem.isAtTarget())
                 .andThen(Commands.parallel(s.indexerSubsystem.runIndexer())))
```
jamesdooley4: Likely should use the new `s.indexerSubsystem.runIndexer(() -> s.flywheels.getTargetSpeed())` so the indexer runs at a speed relative to the flywheel velocity.

jamesdooley4 (dismissed):
Some minor comments to cleanup

src/main/java/frc/robot/subsystems/index/IndexerSubsystem.java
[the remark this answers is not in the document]
markpete: Agree - a comment explaining here would be sufficient.

src/main/java/frc/robot/Controls.java
```diff
@@ -377,13 +376,23 @@ private void configureIntakeBindings() {
...
+                            () ->
+                                Math.hypot(
+                                            s.drivebaseSubsystem.getState()
+                                                .Speeds
+                                                .vxMetersPerSecond,
+                                            s.drivebaseSubsystem.getState()
+                                                .Speeds
+                                                .vyMetersPerSecond)
+                                        * 6
+                                    + 40);
```
jamesdooley4: Would be more encapsulated to pass `s.drivebaseSubsystem.getState().Speed` to smartIntake and let it do the math internally to figure out the intake speed. Change the `DoubleSupplier` to a `Supplier<ChassisSpeeds>`

jamesdooley4: That also de-duplicates the two calls to `s.drivebaseSubsystem.getState()`

src/main/java/frc/robot/subsystems/launcher/Flywheels.java
jamesdooley4: These likely need to change since the units are Amps instead of Volts.

src/main/java/frc/robot/Controls.java
```diff
@@ -188,7 +187,7 @@ private void configureDrivebaseBindings() {
       return;
     }
 
-    readyToShoot = GetTargetFromPose.autoShoot(s.drivebaseSubsystem);
+    // readyToShoot = GetTargetFromPose.autoShoot(s.drivebaseSubsystem);
```
jamesdooley4, line 191: Need to undo this

markpete (approved):

PR #228, opened 2026-04-15
Climb subsystem

Fatblabs:
Pre process: 
1). Drivebase go a little bit farther away from climb (stage 1) so we can get accurate readings from vision and deploy climb
2). Once climb deployed, go to stage 2 (attach to climb). Check accelerometer if we feel a "jerk" or "slam", which would likely indicate that the robot has attached. Apply a motor voltage and check to see if the current spikes and maintains it. If both conditions pass then we have successfully attached to the climb structure

Climb process:
1). You can request the desired climb level by calling goToLevel(ClimbLevel level)

Post process: 
1). if goTolevel is called with level 0, it lowers the robot down then runs the detachClimb() routine
2). detach climb rotates the drive base with the center of rotation being at the magnet climb. In theory, it should just peel off. As of right now there is no check to see if it has detached or not. maybe pose check? 

Code is designed to be tested in stages, so debugging should in theory be much easier


RobototesProgrammers:
this is so over complicated
view https://github.com/robototes/REBUILT2026/tree/MagnetClimb

Fatblabs:
This is designed to offload as much user input as possible

src/main/java/frc/robot/subsystems/Climb/ClimbPIvot.java
```diff
@@ -0,0 +1,42 @@
...
+  public void deploy() {
+    pivotServo.set(SERVO_DEPLOYED);
+  }
+
+  public void stow() {
+    pivotServo.set(SERVO_STOWED);
+  }
+
+  public boolean isDeployed() {
+    return pivotServo.get() == SERVO_DEPLOYED;
```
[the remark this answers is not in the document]
Fatblabs: aarya made this code.

Fatblabs:
Testing: To tune minimum jerk, plot the jerk in advantage scope and go as slow as possible to the climb structure until it attaches. Measure the magnitude of the jerk. This value is going to be the min G's /s. For MIN_ATTACHED_AMPS, attach the robot to the climb structure, then apply a small voltage to the  climb rollers and see the current. If it doesn't move, that current with that applied voltage will be your constants

.vscode/settings.json
```diff
@@ -1,68 +0,0 @@
...
-    "edu.wpi.first.math.**.proto.*",
-    "edu.wpi.first.math.**.struct.*",
-  ],
-  "java.dependency.enableDependencyCheckup": false,
-  "[java]": {
-    "editor.defaultFormatter": "josevseb.google-java-format-for-vs-code"
-  },
-  "java.format.enabled": false,
-  "files.trimTrailingWhitespace": true,
-  "files.insertFinalNewline": true,
```
jamesdooley4, line 66: No more autoformat. :(
A lot of the rest of this file is autogenerated as part of the WPILib project generator, so I suspect it's needed for something important to the WPILib build or project management.

src/main/java/frc/robot/subsystems/Climb/ClimbPivot.java
```diff
@@ -0,0 +1,39 @@
...
+    return Commands.runOnce(this::deploy).withName("Deploy pivot");
+  }
+
+  public Command stowCommand() {
+    return runOnce(this::stow).withName("Stow Pivot");
+  }
+
+  @Override
+  public void periodic() {
+    SmartDashboard.putNumber("Pivot/Servo Position", pivotServo.get());
```
jamesdooley4: FYI that `.get()` just returns the last value passed to `pivotServo.set()` since servo's can't report their position. Nothing to change here, just don't be surprised to see this value  jump between 0.0 and 0.5.

src/main/java/frc/robot/subsystems/Climb/ClimbPivot.java
```diff
@@ -0,0 +1,39 @@
...
+  public Command deployCommand() {
+    return Commands.runOnce(this::deploy).withName("Deploy pivot");
+  }
+
+  public Command stowCommand() {
+    return runOnce(this::stow).withName("Stow Pivot");
+  }
+
+  @Override
+  public void periodic() {
```
markpete: Do these need to be in the periodic?  They would be better tied to the actual commands to keep from using unnecessary noise in the CAN bus.

src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,551 @@
...
+  }
+
+  // -------- SENSOR CHECKS -------- //
+
+  public boolean passedAccelerometerTest() {
+    double jerk = getJerk();
+    return jerk >= MIN_JERK;
+  }
+
+  public double getJerk() {
```
markpete: ### Stateful Derivative getJerk() Returns 0 During Climb Loop
src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java Lines: 256-279 (getJerk logic), 544-545 (Periodic execution), 251-254 (Command execution)

Description: The getJerk() method relies heavily on mutation of internal states (lastTime, lastXAccel etc.) to measure the delta-time change. Because you call getJerk() in periodic() to observe telemetry, the lastTime refreshes identically to the robot sequence. But when the attachToClimb() command runs, it inherently calls passedAccelerometerTest() inside its execute block right after periodic() finishes—meaning the computed currentDt is mere microseconds (near identical). Therefore, currentDt >= MIN_JERK_DT evaluates to false, jumping right into discarding the jerk output entirely and returning 0. Consequently, passedAccelerometerTest() will essentially never return true, meaning the system hangs forever trying to climb.
Suggestion: Make getJerk() a private updateJerk() void method that is exclusively called once per loop in periodic(). Assign the result to a private field variable (e.g. this.currentJerkData). Then, change passedAccelerometerTest() to return this.currentJerkData >= MIN_JERK purely reading the recorded data without mutating any state variables.

src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,551 @@
...
+              return Commands.sequence(
+                  Commands.parallel(
+                      new AutoAlignCommand(getStage1Pose(), Translation2d.kZero),
+                      climbPivotSubsystem
+                          .deployCommand()
+                          .andThen(Commands.waitSeconds(ClimbPivot.PIVOT_DELAY_SECONDS))),
+                  Commands.parallel(
+                          new AutoAlignCommand(getStage2Pose(), Translation2d.kZero),
+                          Commands.sequence(
+                                  Commands.runOnce(
```
markpete, line 393: **Motor Runs Endlessly on Loop Timeouts**
src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java Lines: 383-398

Description: Within attachToClimb(), the attempts sequence assigns .withTimeout(ATTACH_TIMEOUT) on the block containing setVoltage(MAX_VOLTS). If this loop times out because passedAccelerometerTest isn't satiated fast enough, it forcibly exits the timeout boundary. However, there is no termination logic configured closely attached to that block—so the TalonFX is left powered on at MAX_VOLTS during the wrap-around iteration while it prepares the next sequence alignment attempt.
Suggestion: Cleanly wrap the specific attempt loop with a halt feature so the motor definitely halts immediately if a timeout occurs before the next .run(...) iteration begins: Fix: Append .finallyDo((interrupted) -> setVoltage(0)) directly before .onlyWhile(...) on line 397.

src/main/java/frc/robot/subsystems/Climb/ClimbPivot.java
```diff
@@ -0,0 +1,39 @@
...
+    pivotServo.set(SERVO_DEPLOYED);
+  }
+
+  public void stow() {
+    pivotServo.set(SERVO_STOWED);
+  }
+
+  public Command deployCommand() {
+    return Commands.runOnce(this::deploy).withName("Deploy pivot");
+  }
```
markpete, line 41: **Missing ClimbPivot Scheduler Subsystem Requirements**
src/main/java/frc/robot/subsystems/Climb/ClimbPivot.java Lines: 26-32

Description: The methods deployCommand() and stowCommand() map to Commands.runOnce(this::deploy). Because this is not supplied as an explicit requirement flag to WPILib's .runOnce, the returned commands do not "own" the ClimbPivot subsystem in WPILib's internal scheduler. If someone manually invokes another pivot feature in TeleOp while the attachToClimb parallel sequence invokes .deployCommand(), WPILib will not interrupt them successfully.
Suggestion: Map requirements appropriately to all exposed subsystem commands. Fix: return Commands.runOnce(this::deploy, this).withName("Deploy pivot");

src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,562 @@
...
+  private static final double MAX_ROTATIONAL_POWER = 4.0;
+  private static final double TRANSLATION_TOLERANCE_METERS = 0.03;
+  private static final double ROTATION_TOLERANCE_RADIANS = Units.degreesToRadians(0.5);
+
+  // -------- MOTOR TUNABLES -------- //
+
+  // Motor Tunables
+  private static final double KP = 3;
+  private static final double KI = 0;
+  private static final double KD = 3;
```
jamesdooley4: kD of 3 is huge. It should be at most 1/10 of the kP value

src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,562 @@
...
+  private static final double KS = 3;
```
jamesdooley4: kS of 3 is also huge. It means the motor can't move unless it's given at least 3v. Something like 0.3v seems more reasonable as a starting point to tune from

src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,562 @@
...
+
+  // Maps
+  private final EnumMap<ClimbLevel, Double> climbLevels = new EnumMap<>(ClimbLevel.class);
+
+  // -------- MOTOR REQUESTS -------- //
+
+  // Motor requests
+  private final MotionMagicVoltage mmRequest;
+  private final VoltageOut voltageRequest;
+  private static final double MAX_VOLTS = 3;
```
jamesdooley4: Recommend something like 8 to start. 3V is unlikely to do much.

src/main/java/frc/robot/subsystems/Climb/ClimbSubsystem.java
```diff
@@ -0,0 +1,562 @@
...
+                  Commands.parallel(
+                          new AutoAlignCommand(getStage2Pose(), Translation2d.kZero),
+                          Commands.sequence(
+                                  Commands.runOnce(
+                                      () -> {
+                                        // Reset any ongoing timers and use a new debouncer each
+                                        // attempt
+                                        hasMaintainedCurrent =
+                                            new Debouncer(DEBOUNCE_TIME, DEBOUNCE_TYPE);
+                                        setVoltage(MAX_VOLTS);
```
jamesdooley4, line 399: Should use something other than the maximum output. Recommend something like ATTACH_DETECTION_VOLTS and use something like 3.0

src/main/java/frc/robot/Hardware.java
```diff
@@ -19,6 +19,8 @@ public class Hardware {
 
   // Climb Motor (35-39)
   public static final int CLIMB_MOTOR_ID = 35;
+  // Climb Pivot
+  public static final int CLIMB_PIVOT_SERVO_CHANNEL = 0; // PMW port
```
jamesdooley4: Typo: PWM

jamesdooley4 (changes requested):
Good start. Please keep the branch up-to-date with main so it's ready to test when/if climb is physically re-attached.

PR #230, opened 2026-04-16
Acceleration with SOTM

Fatblabs:
Instead of assuming constant velocity rigid body using currentPose.exp() (which assumes the robot will travel in a perfect circular arc in PHASE_DELAY seconds), we account in for acceleration and replace the .exp and apply kinematics transforms directly into currentPose;

Delta (x/y/z) = (vi * t) + (1/2 * a * t^2)

Acceleration calculation: Use a Kalman Filter to feed in position and velocity data to spit out a predicted acceleration.


src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
```diff
@@ -233,7 +266,41 @@ public void periodic() {
...
+    var pose = getState().Pose; // field relative
+    var speeds = getState().Speeds; // robot relative
+
+    ChassisSpeeds fieldSpeeds =
+        ChassisSpeeds.fromRobotRelativeSpeeds(speeds, pose.getRotation()); // field relative
+
+    filterX.update(pose.getX(), fieldSpeeds.vxMetersPerSecond);
+    filterY.update(pose.getY(), fieldSpeeds.vyMetersPerSecond);
+    filterTheta.update(pose.getRotation().getRadians(), fieldSpeeds.omegaRadiansPerSecond);
+
```
markpete, line 278: ### Massive Acceleration Spikes Over Pose Discontinuities
**Severity:** High
**Files & Line Numbers:** 
- `src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java` (Lines 269-279)
- `src/main/java/frc/robot/util/KinematicFilter.java` (Lines 64-82)

**Description & Logic Concern:**
`CommandSwerveDrivetrain.java` fetches the absolute field-relative state (`pose.getX()`, `pose.getY()`) and dynamically pushes it to the `KinematicFilter` inside the 50Hz `periodic()` loop (Lines 275-277).

However, `clampPoseToField()` (Line 279) deliberately overrides and issues instant position corrections if the robot exceeds field bounds through `resetPose`. Furthermore, manual pose resets (e.g. from vision measurements or driver buttons like POV-Left in `Controls.java`) abruptly snap `getState().Pose`.

The continuous state vectors in the WPILib `KalmanFilter` evaluate the rigid discrepancy over `measuredPos`. When an instant position teleport occurs (e.g., snapping from 16.0m back to 15.0m), the filter calculates this differential as massive reversed velocity occurring over `0.02s` alongside the actual logged motor velocity, creating an artificially astronomical acceleration calculation (`ax` or `ay`).

Since these faulty outputs immediately evaluate into `dxField` inside `LaunchCalculator.java` (Line 184), an ill-timed position snapping or vision jump will cause the turret predictor to spin erratically towards an impossible coordinate. 

**Suggested Fixes & Code Fix:**
- **Add an Internal Reset:** Expose a `public void reset(double currentPos)` method inside `KinematicFilter.java`. This should overwrite `accumulatedValue` and `lastValue` to the `currentPos` parameter, while also wiping the internal WPILib `KalmanFilter` state to avoid calculating false velocity/acceleration jumps. Trigger this method when large vision adjustments or `resetPose` occurs.

```java
// Add to src/main/java/frc/robot/util/KinematicFilter.java
public void reset(double currentPos) {
  lastValue = currentPos;
  accumulatedValue = currentPos;
  initialized = true;
  filter.setXhat(0, currentPos);
  filter.setXhat(1, 0.0);
  filter.setXhat(2, 0.0);
}
```

- **Trigger the Reset on Pose Changes:** Override the `resetPose()` mechanism to catch manual origin updates. Additionally, calculate positional deltas in the `periodic()` loop to actively catch sudden aggressive vision snaps that organically bypass the `resetPose()` command.

```java
// Add a state tracker variable to CommandSwerveDrivetrain.java
private Pose2d m_lastPose = new Pose2d();

// Update inside the periodic() method of CommandSwerveDrivetrain.java
Pose2d currentPose = getState().Pose; // field relative

// Check for sudden large vision jumps before updating filters (> 0.5m or > 10 degrees jump in 20ms)
double translationJump = currentPose.getTranslation().getDistance(m_lastPose.getTranslation());
double rotationJump = Math.abs(currentPose.getRotation().minus(m_lastPose.getRotation()).getDegrees());

if (translationJump > 0.5 || rotationJump > 10.0) {
  if (filterX != null) filterX.reset(currentPose.getX());
  if (filterY != null) filterY.reset(currentPose.getY());
  if (filterTheta != null) filterTheta.reset(currentPose.getRotation().getRadians());
}
m_lastPose = currentPose;

// ...proceed with updating the filter...
```

```java
// Update in src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
@Override
public void resetPose(Pose2d pose) {
  super.resetPose(pose);
  if (filterX != null) filterX.reset(pose.getX());
  if (filterY != null) filterY.reset(pose.getY());
  if (filterTheta != null) filterTheta.reset(pose.getRotation().getRadians());
}
```

- **Physical Clamping Safeguards:** Before sending the filtered outputs to standard processing logic, clamp the outputs returned by `filterX.getAccel()` and `filterY.getAccel()` out to physically realistic swerve capabilities. For example: `MathUtil.clamp(val, -15.0, 15.0)`.

```java
// Update in src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
public double getFieldRelativeXAccel() {
  return MathUtil.clamp(filterX.getAccel(), -15.0, 15.0);
}

public double getFieldRelativeYAccel() {
  return MathUtil.clamp(filterY.getAccel(), -15.0, 15.0);
}

public double getAngularAcceleration() {
  return MathUtil.clamp(filterTheta.getAccel(), -25.0, 25.0);
}
```

---

src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
```diff
@@ -233,7 +266,41 @@ public void periodic() {
...
     clampPoseToField();
+
+    // Publish filtered accelerations computed by KinematicFilter
+    try {
```
markpete: The try/catch here isn't really needed.  The null checks should be sufficient.  The only way this would fail is if the network table instantiation silently collapsed, which is unlikely.

markpete (changes requested):

src/main/java/frc/robot/subsystems/drivebase/CommandSwerveDrivetrain.java
Fatblabs: Could I just call the new KinematicFilter's reset() inside CommandSwerveDriveTrain's clampPoseToField()? Becuase if the pose teleports it would most likely always be user activated

jamesdooley4:
@Fatblabs comments from Claude:

# KinematicFilter Usage Review

**Scope:** `frc.robot.util.KinematicFilter` and its use in
`frc.robot.subsystems.drivebase.CommandSwerveDrivetrain`, consumed by
`frc.robot.subsystems.launcher.LaunchCalculator`.

**Symptom being investigated:** After the robot stops, the turret oscillates
back and forth by ~5–20°. It appears worse when no AprilTags are in view, but
this correlation is uncertain.

---

## How the filter is currently wired

`KinematicFilter` is a 3‑state Kalman filter over
`x = [position, velocity, acceleration]ᵀ` with a constant‑acceleration process
model (`A[2,2] = 0`, so `ȧ` is pure process noise). Two measurements are fed
in each `periodic()` call:

| Measurement | Source |
|---|---|
| position | `getState().Pose.getX() / getY() / getRotation().getRadians()` |
| velocity | field‑converted `getState().Speeds` (vx, vy, ω) |

The filter output `getAccel()` returns `x̂[2]` directly with no smoothing and
is then clamped to ±15 m/s² / ±25 rad/s² in the drivetrain getters. No
deadband is applied.

Filter constants in `CommandSwerveDrivetrain`:

```java
new KinematicFilter(0.005, 0.05, 0.5, 0.02, false); // X, Y
new KinematicFilter(0.005, 0.05, 0.2, 0.02, true);  // θ
// (posStdDev, velStdDev, accelProcessStdDev, dt, isAngle)
```

---

## Findings (ordered by likely contribution to the turret vibration)

### 1. No zero‑gate when the robot is stationary  *(most likely root cause)*

When the chassis stops, `fieldSpeeds` does **not** cleanly reach zero —
module encoder quantization, kinematics skew, and pose‑estimator
micro‑corrections keep injecting small correlated residuals. A constant‑
acceleration Kalman filter interprets those residuals as real non‑zero `a`.
The estimate ping‑pongs around zero by tenths of a m/s² and tenths of a
rad/s², which the turret feed‑forward multiplies into several degrees of
command jitter — exactly matching the observed 5–20° vibration.

`CommandSwerveDrivetrain.isStationary()` already exists but is not used to
gate the accel outputs.

**Fix:** force `getAccel()` to `0` when `isStationary()` is true (or when
the speeds are each below a small threshold), e.g.

```java
public double getFieldRelativeXAccel() {
    if (isStationary()) return 0.0;
    return MathUtil.clamp(filterX.getAccel(), -15.0, 15.0);
}
```

This change alone should eliminate the visible buzz.

### 2. Measurement / process noise tuning is too trusting

- `posStdDev = 0.005 m` (5 mm!) and `velStdDev = 0.05 m/s` tell the filter
  to almost fully believe each measurement.
- `accelProcessStdDev = 0.5 m/s² / step` (and `0.2 rad/s² / step`) tells it
  that acceleration can change freely between steps.

The combination produces a filter that faithfully *tracks noise* into the
acceleration state. Starting point for re‑tuning:

| Axis | posStdDev | velStdDev | accelProcessStdDev |
|---|---|---|---|
| X, Y | 0.02 m  | 0.15 m/s  | 0.2 m/s² |
| θ    | 0.02 rad | 0.10 rad/s | 0.1 rad/s² |

The cost is a small amount of phase lag, which is acceptable for launcher
feed‑forward.

### 3. Position and velocity measurements are not independent

`Speeds` is derived by differencing module positions — the same source as
`Pose`. Feeding *both* lies to the Kalman filter about how much independent
information it has, which effectively shrinks `R` and amplifies Finding 2.

Cleaner options:

- **Position‑only** filter: `C = [1 0 0]`, let the filter derive v and a.
- **Velocity‑only** filter: drop position entirely, run a 2‑state `[v, a]`
  filter on `Speeds`. This is the most robust choice and also sidesteps
  Finding 5 below.

### 4. `dt` is hard‑coded at 0.02 s

`filter.predict(..., dt)` is called with a constant `0.02` and the
continuous‑to‑discrete Kalman gains are computed once at construction with
the same `dt`. Odometry samples in `getState()` arrive at 250 Hz (CAN FD)
or 100 Hz (CAN 2.0), and `periodic()` can occasionally overrun. A fixed
`dt` mismatch slowly biases the estimator.

**Fix:** measure the real elapsed time (`Timer.getFPGATimestamp()` delta)
and pass it to `predict(...)`, or drive the filter off the odometry
thread’s real sample rate.

### 5. Sub‑threshold vision/pose‑estimator snaps leak into acceleration

`KinematicFilter.update()` only treats a position jump as a reset when it
exceeds `MIN_JUMP_METERS = 0.5 m` or `MIN_JUMP_DEGREES = 10°`. Typical
AprilTag corrections are 1–5 cm and 1–3°, which sail right under that
threshold. For several ticks after a snap, position moves while velocity
doesn’t, and the filter dumps the mismatch into `a`.

The user’s observation that vibration seems worse *without* AprilTags is
consistent with Finding 1 dominating today, but this mechanism will still
contribute spikes whenever vision updates arrive.

**Fix options:**

- Lower the jump thresholds (e.g. 0.05 m, 2°).
- Switch to a velocity‑only filter (Finding 3) — vision does not touch
  velocity.

### 6. `clampPoseToField()` runs *after* the filter update

When the clamp fires it calls `resetPose(...)`, producing a position
discontinuity that the filter never sees (because the next cycle’s jump
test runs against `lastValue` captured *before* the clamp). If the clamp
is < 0.5 m the filter will integrate a fake acceleration spike over
several cycles.

**Fix:** either perform the clamp *before* the `filter*.update(...)`
calls, or call `filterX.reset(...)` / `filterY.reset(...)` from inside
`clampPoseToField()` when it actually modifies the pose.

### 7. Minor: angle accumulation uses wrapped `lastValue` for jump detection

The jump check uses `MathUtil.angleModulus(measuredPos - lastValue)` with
`lastValue` storing the most recent *wrapped* angle. That is correct but
fragile — an extremely fast real yaw rate (~30 rad/s, not physically
achievable) could alias into a false reset. Not a current bug, just noted
for awareness.

### 8. No output smoothing

Even after the changes above, a light one‑pole LP (`α ≈ 0.2`) or
`SlewRateLimiter` on the values returned from the three accel getters
would cost ~10 ms of extra lag and visibly remove residual buzz. Optional.

---

## Recommended fix order (least risky → most intrusive)

1. **Zero‑gate when stationary** in the three `getFieldRelative*Accel()`
   getters (Finding 1). Expected to resolve the reported vibration by
   itself.
2. **Re‑tune noise parameters** (Finding 2) to the table above and verify
   via the already‑published `/DriveState/Accelerations/filtered*` NT
   topics against logged real motion.
3. **Fix the clamp ordering** so `clampPoseToField()` runs before
   `filter*.update(...)`, or make it reset the filters (Finding 6).
4. **Use real elapsed `dt`** in `predict(...)` (Finding 4).
5. **Refactor to a velocity‑only 2‑state filter** fed by `Speeds`
   (Findings 3 & 5). Biggest change, biggest robustness win — do in a
   dedicated PR once the cheaper fixes are validated.
6. *(Optional)* Add a light LP on the accel outputs (Finding 8).

---

## Consumer context

`LaunchCalculator.getParameters(...)` already caches results when speeds
and rotation are below small tolerances, but it still calls
`calculate(...)` with live `getFieldRelativeXAccel / YAccel /
AngularAcceleration` whenever those cache conditions aren’t all met. The
stationary‑gate in Finding 1 is therefore load‑bearing for the moment the
robot *just* stopped but before every cache condition has settled — which
is exactly when the reported turret twitch occurs.

PR #231, opened 2026-04-17
Magenticlimitswitch

Jetblackdragon:


Fatblabs (dismissed):
Idk if my input means anything but seems good to me.

AprilTagPipelines/AprilTagA(1).vpr
markpete, line 1: Are you happy with the file naming w/ "(1)"?  Seems like an error.

src/main/java/frc/robot/subsystems/launcher/TurretSubsystem.java
```diff
@@ -309,15 +311,7 @@ public Command voltageControl(Supplier<Voltage> voltageSupplier) {
         .withName("Voltage Control");
   }
 
-  public Command autoZeroCommand() {
```
markpete, line 312: I assume removing the autoZeroCommand is intentional, but I'm not sure what it has to do with the limit switch.  OK, for now, but future best practice would be for this to be it's own PR.

src/main/java/frc/robot/subsystems/VisionSubsystem.java
markpete, line 148: Fine for now, an off-season task might be changing this to a dynamic array or Collection of instances, so it's easy to add/remove cameras in the future. You can then set the unique values for each camera and access and process them dynamically.

```
import edu.wpi.first.math.geometry.Pose3d;
import edu.wpi.first.math.geometry.Transform3d;

public class PoseData {
    public Pose3d pose;
    public String label;
    public Transform3d transform;

    // Constructor
    public PoseData(Pose3d pose, String label, Transform3d transform) {
        this.pose = pose;
        this.label = label;
        this.transform = transform;
    }
}
```

You could then also handle your camera processing using a pipeline model for easier codability and understanding.

src/main/java/frc/robot/Controls.java
```diff
@@ -474,7 +473,13 @@ private void configureTurretBindings() {
       return;
     }
 
-    s.turretSubsystem.setDefaultCommand(s.turretSubsystem.rotateToTargetWithCalc());
+    s.turretSubsystem.setDefaultCommand(
+        s.turretSubsystem.rotateToTargetWithCalc().withName("Turret Default Command"));
+    new Trigger(() -> s.turretSubsystem.atLimitSwitch())
+        .onTrue(
+            Commands.runOnce(() -> s.turretSubsystem.zeroTurret())
```
markpete, line 480: ### Turret Zeroing Logic Failure inside Commands
src/main/java/frc/robot/Controls.java Line(s): 480 Severity: High

Description: In Controls.java, the turret limit switch is wired to zero the turret using Commands.runOnce(() -> s.turretSubsystem.zeroTurret()). However, the method zeroTurret() itself returns a Command object. Wrapping it in a runOnce lambda simply executes the method and creates a Command instance, then immediately discards it without actually scheduling it. The internal action of zeroTurret() (e.g. Setting encoder position to 0) will never execute when the hall effect sensor triggers.

Recommendation: Pass the returned command directly to .onTrue() instead of wrapping it in Commands.runOnce:

```
new Trigger(() -> s.turretSubsystem.atLimitSwitch())
    .onTrue(s.turretSubsystem.zeroTurret().withName("Zero Turret on Limit Switch"));
```

markpete (changes requested):
Please fix the commanding issue and then I'll approve.

AprilTagPipelines/AprilTagA(1).vpr
koolpoolo: yeah it's fine

src/main/java/frc/robot/subsystems/VisionSubsystem.java
koolpoolo: we also don't need to do this right now and we might have more important things to do this offseason like systemcore

PR #234, opened 2026-04-20
fixed event markers and auto commands

### LINKED ISSUE #233: Can't use Commands and Event Triggers in Pathplanner *Paths*
This is  great bug that Connor B. (Ben10) found!

It turns out that Event triggers CAN'T be used in Pathplanner Paths (and Commands can't be used either from Paths).

The key bug is this:
In AutoLogic, the paths are automatically loaded in the static initialization of the AutoLogic class variables, e.g. where rebuiltPaths static variable is initialized.

The problem is that then the auto is LOADED in the static initialization when this is called:
     namesToAuto.put(auto.**getDisplayName()**, auto);

**Later**, in Robot class constructor, the commands are register with this call:
AutoLogic.registerCommands(true);

So that means the ordering is:
1) Load paths
2) Register commands

But when the paths are loaded that refer to a command, those commands hadn't been register yet, so they silently fail.

The fix is to change the ordering to:
1) Register commands
2) Load paths

Here's an example Commit of how it can be fixed:
[a34c749feaa55060ef277a2b51906e9b73265eb5
](https://github.com/robototes/REBUILT2026/commit/a34c749feaa55060ef277a2b51906e9b73265eb5)

Since this changes ordering of initialization, it is medium-risk.  So probably a change that could happen after Worlds.  Depends on how badly you want to run EventTriggers during auto?

koolpoolo:


src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -110,7 +90,68 @@ public static void init(Subsystems subsystems) {
     s = subsystems;
   }
 
+  // We always need to register commands BEFORE we initialize paths.  This is
+  // because paths may reference commands or triggers that need to be registered first.
+  // This helper-method insures caller initializes these in the correct order.
+  public static void initCommandsAndPaths(boolean testMode) {
```
markpete: (Optional change, you can ignore this one if you like.)

You can remove the overload using Optional

```
import Java.util.Optional;
public static void initCommandsAndPath(Optional<boolean> testMode) {
   if (!testMode.orElse(false) {
      registerCommands(true);
   }
   initPaths();
}
```

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -110,7 +90,68 @@ public static void init(Subsystems subsystems) {
...
+      }
+    }
+
+    pathsInitialized = true;
+  }
+
+  private static void requirePathsInitialized() {
+    if (!pathsInitialized) {
+      throw new IllegalStateException(
+          "Auto paths are not initialized. Call AutoLogic.initPaths().");
```
markpete: Description: The IllegalStateException triggered in requirePathsInitialized() tells the developer to "Call AutoLogic.initPaths().". However, initPaths() has been made private in this PR. Developers attempting to trace and fix this error will be confused because they cannot invoke that private method directly from external caller classes.
Recommendation: Update the exception text to instruct the caller to use the intended public initialization method: throw new IllegalStateException("Auto paths are not initialized. Call AutoLogic.initCommandsAndPaths().");

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -110,7 +90,68 @@ public static void init(Subsystems subsystems) {
...
+    initPaths();
+  }
+
+  public static void initCommandsAndPaths() {
+    initCommandsAndPaths(false);
+  }
+
+  private static void initPaths() {
+    List<AutoPath> physicalRebuiltPaths;
+    List<AutoPath> simRebuiltPaths;
```
markpete: This variable is never used.

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -185,7 +247,7 @@ public static Command getAutoCommand(String pathName)
     return AutoBuilder.followPath(path);
   }
 
-  public static void registerCommands(boolean enabled) {
+  private static void registerCommands(boolean enabled) {
```
markpete: Description: The registerCommands(boolean enabled) method was successfully refactored from public to private. Because the only internal call to it dictates registerCommands(true) explicitly, the enabled parameter is now redundant. Consequently, the else block branching on enabled == false is entirely dead and unreachable code.
Recommendation: Remove the boolean enabled parameter from the registerCommands() signature, and eliminate the unreachable else block (Lines 266-271) to simplify the class logic.

src/test/java/AutosTest.java
```diff
@@ -23,12 +23,21 @@ void setup() {
...
+
     Dictionary<String, Boolean> pathDictionary = new Hashtable<>();
     for (AutoPath path : AutoLogic.getAutos()) { // Every auto in the auto chooser
       pathDictionary.put(
           path.getAutoName(), false); // Putting each auto from the auto chooser in dictionary
     }
 
+    // We expect >= 1 paths, otherwise its possible we forgot to call
+    // AutoLogic::initCommandsAndPaths().
+    if (pathDictionary.size() == 0) {
```
markpete: Description: The test conditionally checks if there are items in the dictionary using pathDictionary.size() == 0. The standard Java idiom for checking if a collection or map is empty is to use the native .isEmpty() method which reads much clearer.
Recommendation: Refactor the condition to if (pathDictionary.isEmpty()) {.

markpete (changes requested):

src/test/java/AutosTest.java
koolpoolo: done

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
koolpoolo: done

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
koolpoolo: deleted

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
koolpoolo: done

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
koolpoolo: done

src/main/java/frc/robot/subsystems/auto/AutoLogic.java
```diff
@@ -110,7 +91,67 @@ public static void init(Subsystems subsystems) {
     s = subsystems;
   }
 
+  // We always need to register commands BEFORE we initialize paths.  This is
+  // because paths may reference commands or triggers that need to be registered first.
+  // This helper-method insures caller initializes these in the correct order.
+  public static void initCommandsAndPaths(Optional<Boolean> testMode) {
+    if (testMode == null || !testMode.orElse(false)) {
```
markpete: if (testMode != null && testMode.orElse(false))

markpete (approved):

PR #235, opened 2026-04-21
GC stats

Roxie222:


src/main/java/frc/robot/util/GCMonitor.java
```diff
@@ -0,0 +1,79 @@
...
+ * counter each time a GC occurs. It also writes the count to SmartDashboard under the key "GCCount"
+ * and logs an info line.
+ *
+ * <p>Enable by calling {@link #start()} early in application startup (Robot constructor).
+ */
+public final class GCMonitor {
+  private static final AtomicLong gcCount = new AtomicLong(0);
+  private static volatile boolean started = false;
+
+  private GCMonitor() {}
```
markpete, line 24: Class constructors should not have a private/public prefix.

src/main/java/frc/robot/Robot.java
```diff
@@ -87,6 +87,8 @@ protected Robot() {
     LiveWindow.disableAllTelemetry();
     LiveWindow.enableTelemetry(PDH);
     BuildInfo.logBuildInfo();
+    // Start GC monitor to count garbage collections and publish to SmartDashboard
+    frc.robot.util.GCMonitor.start();
```
markpete, line 92: Should this be behind a debug flag?  I don't think we want this running in competition.

src/main/java/frc/robot/util/GCMonitor.java
```diff
@@ -0,0 +1,79 @@
...
+                      DataLogManager.log(
+                          String.format(
+                              "GC #%d action=%s name=%s duration=%dms",
+                              current,
+                              info.getGcAction(),
+                              info.getGcName(),
+                              info.getGcInfo().getDuration()));
+                      try {
+                        SmartDashboard.putNumber("GCCount", current);
+                      } catch (Throwable t) {
```
markpete: File: src/main/java/frc/robot/util/GCMonitor.java Lines: 57, 60, 71 Severity: Medium

Description: The code uses catch (Throwable t) in three different try-catch blocks. Catching Throwable is strongly discouraged because it indiscriminately catches irrecoverable Java Virtual Machine errors (e.g., OutOfMemoryError, StackOverflowError, ThreadDeath). Swallowing these critical errors, especially within background execution threads like a JMX notification handler, can obscure serious systemic failure during a match and make debugging almost impossible.

Recommendation: Change catch (Throwable t) to catch (Exception e). This allows you to safely handle routine runtime exceptions in WPILib/SmartDashboard or JMX operations while allowing fatal JVM errors to properly bubble up and crash the process cleanly, as intended.

src/main/java/frc/robot/util/GCMonitor.java
```diff
@@ -0,0 +1,79 @@
...
+                      // WPILib tools
+                      DataLogManager.log(
+                          String.format(
+                              "GC #%d action=%s name=%s duration=%dms",
+                              current,
+                              info.getGcAction(),
+                              info.getGcName(),
+                              info.getGcInfo().getDuration()));
+                      try {
+                        SmartDashboard.putNumber("GCCount", current);
```
markpete: File: src/main/java/frc/robot/util/GCMonitor.java Line: 56 (SmartDashboard.putNumber("GCCount", current);) Severity: Medium

Description: SmartDashboard.putNumber is being executed inside the handleNotification method, which is invoked by an internal JMX background thread whenever a Garbage Collection event fires. While WPILib's NetworkTables 4 is highly lock-free and thread-safe, it is a known FRC best practice to avoid tightly coupling internal JVM thread behavior with the NetworkTables transmission queue. Doing so runs the risk of stalling the JMX notification thread or queuing operations arbitrarily, which can cause unpredictable timing behavior.

Recommendation: Remove the SmartDashboard.putNumber call from the handleNotification callback. Since the class already statically hosts the updated value in gcCount, the recommended FRC pattern is to read GCMonitor.getGcCount() in your main loop (specifically Robot.robotPeriodic()) and publish the value to SmartDashboard there. Note: Using DataLogManager.log from the background thread here is acceptable, as it guarantees asynchronous queueing under the hood.

markpete (dismissed):

src/main/java/frc/robot/Robot.java
jamesdooley4: We do - this is to debug all the loop overruns we keep getting.

src/main/java/frc/robot/util/GCMonitor.java
jamesdooley4: This is intentional since this class shouldn't ever be created - it only has static methods.

jamesdooley4 (approved):
Verified working on robot

RobototesProgrammers (approved):

PR #240, opened 2026-04-21
sotm with acceleration

Fatblabs:
This one uses pigeon 2 rather than chassispeeds


SnappyRiffs:
wait why is there 2 prs with acceleration w/ sotm

PR #242, opened 2026-04-22
4/22 dp

RobototesProgrammers:
I just had claude make changes based on when it wouldnt work


RobototesProgrammers:
I think this is similar to the sotm-acceleration branch idk

Jetblackdragon:
this is now dp changes

PR #258, opened 2026-05-02
shotBlocker in code

cwkpete7383:


koolpoolo:
seems like mech is doing this so keeping this here just in case

PR #262, opened 2026-05-31
Fix BLine chained path pose resets

edanliahovetsky:
## Summary
- avoid resetting robot rotation from `getInitialModuleDirection()` when building BLine paths
- use a pose-resetting builder for the first path and a non-resetting builder for continuation paths
- route chained autos through `buildPath(path, resetPose)` so only the first path resets odometry

## Root cause
`getInitialModuleDirection()` is a module/wheel direction hint, not the robot odometry heading. Resetting robot rotation to it between paths caused heading snaps. The existing `FollowPath.Builder` also retained `.withPoseReset(...)`, so every later `build(path)` reset pose again.

## Validation
- `./gradlew compileJava -PautomatedTest=true --console=plain`
- `./gradlew test -PautomatedTest=true --console=plain`
- Simulated BLine chained autos with wpilib-agent-tools and inspected WPILOGs in AdvantageScope

## Scope
This PR only changes `src/main/java/frc/robot/subsystems/auto/BLine/BLineLogic.java`. No path JSON, test harness, or generated log files are included.

### DIFF
--- a/src/main/java/frc/robot/subsystems/auto/BLine/BLineLogic.java
+++ b/src/main/java/frc/robot/subsystems/auto/BLine/BLineLogic.java
@@ -66,6 +66,7 @@ public class BLineLogic {
   private static boolean pathsInitialized = false;
 
   public static FollowPath.Builder pathBuilder;
+  private static FollowPath.Builder continuingPathBuilder;
   private static Command bLineLaunching;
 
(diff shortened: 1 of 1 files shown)


SnappleRamen:
Added these fixes into the BLineFallback branch

PR #268, opened 2026-08-14
Updated Gemini Automatic Code Review Agent

markpete:
Google is ending support for their consumer version of Google Code Assist.  According to my research this change MAY still work and enable automatic as well as trigger (by /gemini command) code reviews in our repository.

Unfortunately, this needs to be checked into main to see if it works...


.github/workflows/gemini-review.yml
```diff
@@ -0,0 +1,186 @@
...
+
+          echo "PR_NUMBER=$PR_NUM" >> $GITHUB_ENV
+
+          # Get PR Head Commit SHA (required for inline comments)
+          COMMIT_SHA=$(gh pr view "$PR_NUM" --repo "${{ github.repository }}" --json headRefOid -q .headRefOid)
+          echo "COMMIT_SHA=$COMMIT_SHA" >> $GITHUB_ENV
+
+          # Download PR diff
+          gh pr diff "$PR_NUM" --repo "${{ github.repository }}" > pr_diff.txt
+          head -c 20000 pr_diff.txt > trimmed_diff.txt
```
koolpoolo: make the limit 200000 instead of 20000 because diffs will be larger than 20000 characters

koolpoolo (changes requested):

.github/workflows/gemini-review.yml
markpete: That would be a monstrous PR.  The whole point here is to limit cost overages from giant PR's.

koolpoolo (approved):

PR #269, opened 2026-08-14
Updated spotless to run on the latest version of Node.js to reduce output errors

markpete:
When spotless currently runs it complains that it's running on an old version of Node.js.  This change updates to the latest code which runs on a current version of Node.js and reduces output errors for PR's.


koolpoolo (approved):
seems good to me

